PR #5, opened 2023-03-02
spec seeding

adurbin-rivos:
- Change the resulting pdf name to riscv-brs-spec.pdf
- Use brs.bib as the bibtex file instead of example.bib
- Ignore vim swap files.
- Change bibtex format to IEEE to make citings smaller.
- Seed Glossary and Bibliography with some expected terms.
- Add BRS-I and BRS-B names
- Add RISC-V Profile to bibliography & glossary
- Add dependencies in table form to BRS-I and BRS-B recipes
- Add initial introduction stab.


andreiw:
looks good to me

PR #25, opened 2023-03-30
acpi: flesh out and rework compared to the illustrative doc

andreiw:
The spec aims to provide only the additional BRS requirements on top of existing specification and industry norms.

### DIFF
--- a/acpi.adoc
+++ b/acpi.adoc
@@ -7,7 +7,7 @@ This section defines mandatory and optional ACPI requirements on top of the ACPI
 
 === ACPI High-Level Requirements
 
-A compliant platform must:
+A compliant system must:
 
 * Conform to UEFI specification requirements for ACPI:
 ** RISC-V Platforms (cite:[UEFI] Section 2.3.7)
@@ -28,7 +28,7 @@ ACPI support.
 
 ==== HW-Reduced ACPI Model
 
-Compliant RISC-V platforms only implement the HW-Reduced ACPI Model. This means the hardware portion of
+Compliant RISC-V systems only implement the HW-Reduced ACPI Model. This means the hardware portion of
 cite:[ACPI] (Section 4) is not required or supported. All functionality is instead provided through equivalent
 software-defined interfaces and the complexity in supporting ACPI is reduced.
 
@@ -44,7 +44,7 @@ The following additional BRS requirements apply:
 
 ==== Hart Performance Control
 
-If the platform provides support for OS-directed hart performance control, then it must be exposed using
+If the system provides support for OS-directed hart performance control, then it must be exposed using
 Collaborative Processor Performance Control (CPPC, cite:[ACPI] Section 8).
 
 ==== Time and Alarm Device
@@ -64,7 +64,7 @@ through the driver-backed OperationRegion.
 
 This section lists additional BRS-specific requirements
 for the mandatory and conditionally-required ACPI tables in a compliant
-platform.
+system.
 
 This section does not list ACPI tables without additional BRS-imposed
 requirements.
@@ -77,7 +77,7 @@ syntax and semantics of the ACPI or other relevant firmware specification.
 Entry ordering can be correlated with initialization order by an OS, but
 should not be taken to reflect affinity in resource sharing,
 e.g. sockets, caches, etc. RINTC hart ID and ACPI Processor UID should
-not be decoded in a platform-specific manner to divine CPU topology.
+not be decoded in a system-specific manner to divine CPU topology.
 The PPTT Processor Properties Topology Table (PPTT) is to be used to
 describe affinities.
 
@@ -85,13 +85,13 @@ describe affinities.
 
 The Processor Properties Table (PPTT, cite:[ACPI] Section 5.2.29)
 describes the hart affinities and shared resources, such as caches.
-It is mandatory, even for platforms with a simple hart topology.
+It is mandatory, even for systems with a simple hart topology.
 
 ==== Conditionally Required ACPI Tables
 
 This section lists BRS-specific requirements for ACPI tables that depend
-on hardware features that may not be present in compliant platforms.
-For example, some platforms may not have PCIe busses or a serial port.
+on hardware features that may not be present in compliant systems.
+For example, some systems may not have PCIe busses or a serial port.
 
 ===== MCFG
 
--- a/acpi.adoc
+++ b/acpi.adoc
@@ -28,12 +28,14 @@ ACPI support.
 
 ==== HW-Reduced ACPI Model
 
-Compliant RISC-V systems only implement the HW-Reduced ACPI Model. This means the hardware portion of
-cite:[ACPI] (Section 4) is not required or supported. All functionality is instead provided through equivalent
+Compliant RISC-V systems only implement the HW-Reduced ACPI Model cite:[ACPI] (Section 4.1).
+This means the hardware portion of cite:[ACPI] (Section 4) is not required or
+supported. All functionality is instead provided through equivalent
 software-defined interfaces and the complexity in supporting ACPI is reduced.
 
 The following additional BRS requirements apply:
 
+* No FACS table.
 * Both interrupt-signaled and GPIO-signaled ACPI Events are supported
   (cite:[ACPI] Section 5.6)
 * There is at least 1 wake event.
--- a/acpi.adoc
+++ b/acpi.adoc
@@ -1,26 +1,142 @@
 [[acpi]]
 == ACPI Requirements
-=== ACPI Version
-=== ACPI-provided Data Structures
-=== ACPI Tables
-==== Mandatory ACPI Tables
-===== RSDP
-===== XSDT
-===== FADT
-===== DSDT and SSDT
-===== MADT
-===== RHCT
-===== SPCR
-===== MCFG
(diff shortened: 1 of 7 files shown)


acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+
+==== HW-Reduced ACPI Model
+
+Compliant RISC-V platforms only implement the HW-Reduced ACPI Model. This means the hardware portion of
+cite:[ACPI] (Section 4) is not required or supported. All functionality is instead provided through equivalent
+software-defined interfaces and the complexity in supporting ACPI is reduced.
+
+The following additional BRS requirements apply:
+
+* Both interrupt-signaled and GPIO-signaled ACPI Events are supported
```
adurbin-rivos, line 37: Do we need to call this out?

For gpio-signaled events that transitively indicates gpios are a requirement.

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+  (cite:[ACPI] Section 5.6)
+* There is at least 1 wake event.
+** Note: for systems that do not support Sx states except S5 soft off, this can be just the power button.
+* For the ACPI Platform Error Interfaces (APEI) support:
+** There is 1 event for non-fatal error signaling (cite:[ACPI] Section 18.3.2.7.2)
+** There is 1 NMI-equivalent signal for use in fatal errors (cite:[ACPI] Section 18)
```
adurbin-rivos, line 43: Do we not need to go into the definition of event vs signal with these two items?

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+==== Time and Alarm Device
+
+UEFI Runtime Service should be used to provide time services to the
+OS, unless the RTC is subject to arbitration issues between concurrent
+OS and UEFI accesses to the underlying hardware. See <<<UEFI Real-Time
+Clock>>>.
+
+In situations where the TAD device depends on a vendor-specific OS
+driver for correct function (SPI, I2C, etc), the TAD must be
+function if the OS driver is not loaded. E.g. when a dependent
```
adurbin-rivos, line 59: 'must be function'  is this supposed to be be functional?

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+driver is loaded, an AML method switches further accesses to go
+through the driver-backed OperationRegion.
+
+=== Additional ACPI Table Requirements
+
+This section lists additional BRS-specific requirements
+for the mandatory and conditionally-required ACPI tables in a compliant
+platform.
```
adurbin-rivos, line 67: Maybe stay away from 'platform' and use 'system'?

acpi.adoc
andreiw: It needs to be called out because the ACPI spec doesn't call out interrupt-signaled events as a GPE replacement choice. I do have an ASWG ticket for that, so once that is resolved we can remove the text.

Wrt gpios... not necessarily. HW-reduced ACPI supports both interrupt-signaled and GPIO-signaled ACPI events. So if you don't have GPIOS, you can use interrupt signaled (and even if you have GPIOs, you may choose to use interrupt signaled for simple enough use cases). We're providing choice so GPIO is not a requirement for BRS, it /may/ be a requirement depending on how you apply the ACPI spec. It doesn't hurt to call this out, but this might be in the guidance section instead.

acpi.adoc
andreiw: To the extent these are used/defined by the ACPI spec - I think the answer is no. We appropriate reference the ACPI APEI content and understanding that is a prerequisite.

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+Collaborative Processor Performance Control (CPPC, cite:[ACPI] Section 8).
+
+==== Time and Alarm Device
+
+UEFI Runtime Service should be used to provide time services to the
+OS, unless the RTC is subject to arbitration issues between concurrent
+OS and UEFI accesses to the underlying hardware. See <<<UEFI Real-Time
+Clock>>>.
+
+In situations where the TAD device depends on a vendor-specific OS
```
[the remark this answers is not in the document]
andreiw: Good idea.

acpi.adoc
andreiw: Yep.

acpi.adoc
andreiw: Ok. Let's consistently use something. 'System', 'solution' or 'implementation'?

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+
+This section does not list ACPI tables without additional BRS-imposed
+requirements.
+
+All ACPI tables not listed here may be used, if they comply with the
+syntax and semantics of the ACPI or other relevant firmware specification.
+
+==== MADT
+
+Entry ordering can be correlate with initialization order by an OS, but
```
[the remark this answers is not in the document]
andreiw: Yep.

acpi.adoc
adurbin-rivos: sg

acpi.adoc
adurbin-rivos: I was using 'system', but 'implementation' may be better.

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+describe affinities.
+
+==== PPTT
+
+The Processor Properties Table (PPTT, cite:[ACPI] Section 5.2.29)
+describes the hart affinities and shared resources, such as caches.
+It is mandatory, even for platforms with a simple hart topology.
+
+==== Conditionally Required ACPI Tables
+
```
vlsunil, line 91: What about NUMA-related tables like SRAT/SLIT.

acpi.adoc
```diff
@@ -1,26 +1,142 @@
...
+** Note: for systems that do not support Sx states except S5 soft off, this can be just the power button.
+* For the ACPI Platform Error Interfaces (APEI) support:
+** There is 1 event for non-fatal error signaling (cite:[ACPI] Section 18.3.2.7.2)
+** There is 1 NMI-equivalent signal for use in fatal errors (cite:[ACPI] Section 18)
+
+==== Hart Performance Control
+
+If the platform provides support for OS-directed hart performance control, then it must be exposed using
+Collaborative Processor Performance Control (CPPC, cite:[ACPI] Section 8).
+
```
vlsunil, line 49: I understand CPPC is better than P-state. But is there a reason we mandate CPPC? What if the processor can not support fine-grained variations as in CPPC? Also, is there a reason to talk about only performance not power management like LPI?

appendices.adoc
```diff
@@ -26,32 +26,10 @@
...
-==== Time and Alarm
-==== PCI and PCIe
 === CXL Requirements
 ==== CXL Host Bridge
 ==== CXL Root Device
 ===== ACPI Device Object for CXL Root Device
 ==== NUMA
-
+=== Firmware Implementation Guidance
+==== ACPI Implementation Guidance
```
vlsunil, line 35: What is this document? Sorry, I might have missed discussions around this since I cannot attend SEE meetings... But will this doc need to be ratified?

brs.bib
```diff
@@ -3,11 +3,11 @@ @electronic{EBBR
   url = {https://arm-software.github.io/ebbr/}
 }
 @electronic{ACPI,
-  title = {Advanced Configuration and Power Interface Specification},
+  title = {Advanced Configuration and Power Interface Specification 6.6},
   url = {https://uefi.org/specifications}
 }
 @electronic{UEFI,
-  title = {Unified Extensible Firmware Interface Specification},
+  title = {Unified Extensible Firmware Interface Specification 2.11},
```
vlsunil, line 10: Am I correct that 2.11 is used as next UEFI versions due to pending ECRs?

appendices.adoc
adurbin-rivos: What do you mean by 'this document'? The whole thing or this portion? The Guidance sections would be marked non-normative but provide helpful instructions for people implementing BRS.

vlsunil: Ahh OK. I thought Implementation Guidance would be a separate document.

acpi.adoc
vlsunil: For CPPC, should exact section 8.4.6 be given?

acpi.adoc
adurbin-rivos: wdyt think w.r.t. system vs implementation?

andreiw: So I'll leave this out as a separate change since IIUC we're still thinking of what the term should be,

adurbin-rivos: Feel free to pick one and run w/ it.

andreiw: Okay I went with 'system'.

acpi.adoc
vlsunil: Thanks Andrei. I agree with the reasoning and makes sense to mandate CPPC in BRS. LPI text looks fine.

acpi.adoc
vlsunil: Yes, thanks again for helping me to understand the philosophy behind adding things in BRS spec vs Industry standard.

vlsunil (approved):
LGTM

andreiw:
Ok, I think given the feedback so far let's merge what we have.

PR #29, opened 2023-05-01
Fill in minimum hart requirements.

adurbin-rivos:
Seed the minimal hart requirements. Also rename 'processor' to hart
in the filenames and description.


hart.adoc
```diff
@@ -0,0 +1,10 @@
+[[hart]]
+== Hart Requirements
+The BRS specification is minimally prescriptive on the RISC-V hart requirements. It is anticipated that detailed requirements will be driven by target market segment and product/solution requirements. The SEE mandates the minimal set required for a compliant system implementation of the BRS.
+
+The BRS requires harts to least be compliant to:
+
+* RVA20S64, as BRS governs the interactions between 64-bit OS supervisor-mode software and 64-bit firmware, when employed.
+* Sm1p11 or hypervisor (H) extension, to sustain components interfaced via SBI.
```
darius-bluespec, line 8: Isn't this redundant, since the hypervisor extension depends on the privileged extension?  The minimum is just Sm1p11.

adurbin-rivos: Yes,  it could be seen that way.  Couldn't one implement H only w/ an emulated M-mode? I think the original intent was to capture that either just m-mode or environments used w/ VMs (therefore requiring H).

Given that, do you have a suggestion to capture that? Split into 2 bullets?

darius-bluespec: > Couldn't one implement H only w/ an emulated M-mode?

I'm not sure what you mean by "H only".  It is my understanding that the H extension is a strict superset of the S extension.

Supervisor mode software shouldn't be able to tell (or care) whether it is running on a transistors that implement M-mode and the S extension, or the S extension virtualized via the H extension, or the S extension emulated on something else.  Perhaps a non-normative note regarding these possibilities would be helpful, but I don't think the specification needs to mention hypervisor here at all.

adurbin-rivos: OK. I see what you mean. I'll add a note to the intro explaining possible deployments. Then this bullet point would be straight forward given that potential context. Thanks for the feedback.

hart.adoc
```diff
@@ -0,0 +1,10 @@
+[[hart]]
+== Hart Requirements
+The BRS specification is minimally prescriptive on the RISC-V hart requirements. It is anticipated that detailed requirements will be driven by target market segment and product/solution requirements. The SEE mandates the minimal set required for a compliant system implementation of the BRS.
+
+The BRS requires harts to least be compliant to:
+
+* RVA20S64, as BRS governs the interactions between 64-bit OS supervisor-mode software and 64-bit firmware, when employed.
```
andreiw, line 7: Can we cite the UEFI section on RISC-V calling convention as a source of additional RISC-V hart requirements that needs to be referenced by an implementation?

hart.adoc
darius-bluespec: The BRS already says all of that by way of depending on SBI.  Duplicating what is in the SBI specification is not necessary or useful.

hart.adoc
darius-bluespec: This seems confusing and unnecessary.  The BRS depends on the UEFI.  The BRS should not restate UEFI requirements.

andreiw: Not restate. Just point to the section that is of interest. This is what the ACPI section does. The UEFI spec is large enough that referring to the section of interest makes sense.

adurbin-rivos: Which section in UEFI spec? 2.3.7 RISC-V Platforms, but that was before your changes, Andrei. I'll put that there for the time being.

adurbin-rivos: May not be the best place, but I added a note about how to find UEFI expectations related to this.

hart.adoc
adurbin-rivos: I updated the writing.

andreiw (approved):

PR #32, opened 2023-05-04
Expand recipes a bit going into further requirements for BRS-I.

andreiw:
1) A BRS-I compliant system must be able to boot a BRS-I compliant OS. (so long as both target the same BRS-I version).
2) If there are extra features present outside the scope of BRS, these must be exposed/configured out of the box in way that doesn't violate (1).

### DIFF
--- a/non-normative/recipes.adoc
+++ b/non-normative/recipes.adoc
@@ -1,0 +1,32 @@
+[[recipe-brs-i-guidance]]
+===== BRS-I Recipe Guidance
+
+Systems compliant to BRS-I can successfully boot an existing generic
+operating system image without system-specific customizations, yet
+this might result in an unoptimized experience and non-functioning
+I/O devices until further software updates are activated.
+
+The best analogy would be a typical Intel Architecture motherboard from
+the early 2000s: you could install an OS on it, but the built-in graphics
+might be low-resolution and the sound, built-in network port or power
+management might not work out of the box. You could subsequently load
+the right drivers from the media coming with the board or fetch newest
+ones using a well-supported network adapter.
+
+Heterogeneous performance harts (e.g. mix of "performance" and "efficiency"
+harts) is a great example of a feature outside the current scope of BRS-I,
+yet with potential for adverse effects to a generic operating system
+written against the BRS-I spec. Such a feature does not have a standard
+way of being detected by an OS, leading to scheduling anomalies in
+operating systems that are not specially adapted. Consider two identical
+threads that, unbeknownst to the OS, get scheduled on harts with different
+performance characteristics. This might see dramatic difference in forward
+progress being made, with unexpected delays, timeouts or even crashes
+posible. Worse, there could be little to no system messages pointing to
+the cause of such behavior. Thus, such a feature would need to be made
+*opt-in* for BRS-I compliance. This could be done implicitly via loading
+of system-specific drivers, or explicitly via firmware setup and
+configuration utilities. The actual meaning of the default (compatible to
+BRS-I) configuration would be highly specific to the vendor. In this example,
+it could capping all harts at the same performance level, or it could mean
+disabling efficiency harts.
--- a/recipes.adoc
+++ b/recipes.adoc
@@ -11,11 +11,35 @@ and BRS-B (for "Bespoke").
 
 === BRS-I Recipe
 
-To simplify software testing and distribution, many operating system
-providers wish to build a single ACPI-based operating system image
-that can be booted and run on platforms supporting a common
-specification base.  The BRS-I recipe documents a standard runtime
-firmware interface to facilitate this.
+The BRS-I recipe aims to simplify end-user experiences, software
+compatibility and OS distribution, by definining a common specification
+for boot and runtime interfaces. BRS-I is expected to be used by
+general-purpose compute devices such as servers, desktops, laptops and
+other devices with industry expectations on silicon vendor, OS and software
+ecosystem interoperatbility. BRS-I enables operating system
+providers to build a single *generic* operating system image that can be
+*successfully booted* on compliant systems. *Generic* means not requiring
+system-specific customizations - only an implementation of BRS-I
+requirements. *Successfully boot* means basic system configuration,
+sufficient for detecting the need for system-specific drivers and
+loading such drivers.
+
+It is understood that systems will deliver features beyond those covered
+by BRS-I. However, software written against a specific version of BRS-I
+must run, unaltered, without *anomalous and unexpected behavior* on
+systems that include such features and that are compliant to that specific
+version of BRS-I. Such behavior, caused by factors entirely unknown to
+a generic OS, is hard to diagnose and always results in a terrible user
+experience that negatively affects the value of the whole RISC-V
+standards-based ecosystem. *Anomalous and unexpected behavior* is taken
+to mean system instability and worst-case behavior for non-specialized
+workloads, but does not include suboptimal/unoptimized behavior or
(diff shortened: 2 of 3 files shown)


atishp04:
How do we plan to ensure that a system or an OS is compliant ? There were discussion about a compliant test suite in the past. But that never got materialized. Is there such plan associated with the BRS spec as well ?

recipes.adoc
```diff
@@ -11,11 +11,35 @@ and BRS-B (for "Bespoke").
 
 === BRS-I Recipe
 
-To simplify software testing and distribution, many operating system
-providers wish to build a single ACPI-based operating system image
-that can be booted and run on platforms supporting a common
-specification base.  The BRS-I recipe documents a standard runtime
-firmware interface to facilitate this.
+The BRS-I recipe aims to simplify end-user experiences, software
```
atishp04, line 14: end-user experiences - Is that necessary ? Given that the BRS-I cares about successful boot only, the user experience may be completely different one system to another.

atishp04: e.g The graphics/network example you gave above results in completely different end user experience.

non-normative/recipes.adoc
```diff
@@ -0,0 +1,32 @@
...
+performance characteristics. This might see dramatic difference in forward
+progress being made, with unexpected delays, timeouts or even crashes
+posible. Worse, there could be little to no system messages pointing to
+the cause of such behavior. Thus, such a feature would need to be made
+*opt-in* for BRS-I compliance. This could be done implicitly via loading
+of system-specific drivers, or explicitly via firmware setup and
+configuration utilities. The actual meaning of the default (compatible to
+BRS-I) configuration would be highly specific to the vendor. In this example,
+it could capping all harts at the same performance level, or it could mean
+disabling efficiency harts.
```
atishp04, line 32: Does that mean they disable some harts for compliance but enable them in the versions they ship it to the customer ?

non-normative/recipes.adoc
```diff
@@ -0,0 +1,32 @@
...
+yet with potential for adverse effects to a generic operating system
+written against the BRS-I spec. Such a feature does not have a standard
+way of being detected by an OS, leading to scheduling anomalies in
+operating systems that are not specially adapted. Consider two identical
+threads that, unbeknownst to the OS, get scheduled on harts with different
+performance characteristics. This might see dramatic difference in forward
+progress being made, with unexpected delays, timeouts or even crashes
+posible. Worse, there could be little to no system messages pointing to
+the cause of such behavior. Thus, such a feature would need to be made
+*opt-in* for BRS-I compliance. This could be done implicitly via loading
```
atishp04, line 27: Not sure what exactly do you mean by *opt-in*. BRS-I compliance shouldn't include those features at all. Correct ?

recipes.adoc
andreiw: BRS-I simplifies the experience because it makes it trivial to install off-the-shelf OS on a variety of systems, same as is done today for x86 and AArch64 SystemReady SR/ES systems. That's what is meant here.

It's an important goal. If the spec is so weak that it means the same level of "involvement" to figure out how to find/build a bootable/working OS image, we will have failed.

non-normative/recipes.adoc
andreiw: It's an example. All I am saying is that if you have behavior that impacts a generic OS image in the manner defined in this commit, requiring additional software to make things work in a non-anomalous fashion, then such behavior must be only enabled (in whatever fashion) after the supporting software is installed.

For this example, you could:
-  have a UEFI setting that allows disabling E cores (that could be disabled once the relevant non-standard system software is installed)
- cap performance level for all harts

...for any approach taken, the feature could be unlocked via a hardware access, SBI call, UEFI NVRAM variable access, etc from the system software itself (or if you can't - manually in UEFI setup).

Again the point is that if you have a feature that prevents a generic BRS-I compliant OS from booting correctly (within the boundaries identified), then yes your feature is only compliant to BRS-I if it is opt-in instead of forced on every boot.

non-normative/recipes.adoc
andreiw: If your system has a feature that prevents BRS-I operating systems from operating correctly (crashes, hangs, timeouts, etc) unless a custom kernel is installed (that is aware of your system's non-standard feature or of your non-standard way of exposing a feature), then this feature cannot be automatically enabled at boot. That's what this is saying. The big.LITTLE example is a good example of what I mean.

andreiw:
> How do we plan to ensure that a system or an OS is compliant ? There were discussion about a compliant test suite in the past. But that never got materialized. Is there such plan associated with the BRS spec as well ?

There's a compliance test suite in development, to be presented soon in the BRS meeting. This would be used to validate an implementation.

Ensuring an OS is compliant is not part of a plan. Overall that would be fairly hard to accomplish given the potential variance in OS software, workarounds in system software for other systems and product productization priorities. This would be mostly self-certified, but an independent body could perform validation on existing platforms in the relevant market segments and this could be made a prerequisite for any public claims or listing....which only matters if "standardized RISC-V" is treated as a brand, with a logo, logo requirements, and some public web site at least listing the various specs and implementations (hw and sw) compliant to these.

recipes.adoc
atishp04: It makes sense now. I misinterpreted the "end-user" experience earlier. The build/boot experience should be as boring as possible. Anybody with prior experience  with ARM64/x86 should be able to get it working in no time.

non-normative/recipes.adoc
atishp04: Ahh got it. You mean the feature must be enabled at runtime rather than at boot time. In that way, compliance is not affected.

non-normative/recipes.adoc
atishp04: Got it. Any custom feature prevents the platform to boot BRS-I operating systems makes it incompatible.
However, if the platform boots while disabling that feature but enables at runtime, it is called *opt-in* for the sake of compliance.

atishp04:
> 



> > How do we plan to ensure that a system or an OS is compliant ? There were discussion about a compliant test suite in the past. But that never got materialized. Is there such plan associated with the BRS spec as well ?
> 
> There's a compliance test suite in development, to be presented soon in the BRS meeting. This would be used to validate an implementation.

Thanks. Looking forward to the presentation.

> 
> Ensuring an OS is compliant is not part of a plan. Overall that would be fairly hard to accomplish given the potential variance in OS software, workarounds in system software for other systems and product productization priorities. This 
would be mostly self-certified, but an independent body could perform validation on existing platforms in the relevant market segments and this could be made a prerequisite for any public claims or listing....which only matters if "standardized RISC-V" is treated as a brand, with a logo, logo requirements, and some public web site at least listing the various specs and implementations (hw and sw) compliant to these.

Yes. Those were the challenges discussed last time platform spec was being designed. In the end, RVI was not very sure how to create that brand value. I am hoping we can do something like SystemReady in ARM. It's just forcing that in RISC-V ecosystem becomes very hard.

non-normative/recipes.adoc
andreiw: Yes, I think we're on the same page now.

(to be pedantic, the feature *could* be enabled at boot time - it's a choice, for example by way of an OS driver setting a UEFI variable so that subsequent boots see the feature enabled, or an operator being asked to enable something in the firmware setup dialogs... the key aspect is that it's *not* a default configuration... going into UEFI setup and "pick defaults" should place the machine into a state where BRS-I compliant OSes boot)

PR #33, opened 2023-05-16
Eliminate superfluous and redundant content

darius-bluespec:


acpi.adoc
```diff
@@ -38,10 +38,6 @@ software-defined interfaces and the complexity in supporting ACPI is reduced.
 The following additional BRS requirements apply:
 
 * No FACS table.
-* Both interrupt-signaled and GPIO-signaled ACPI Events are supported
```
andreiw, line 41: This is actually not referenced in the spec. There's a bug open for that, but interrupt-signaled events are not documented as a choice for hw-reduced systems.

acpi.adoc
```diff
@@ -38,10 +38,6 @@ software-defined interfaces and the complexity in supporting ACPI is reduced.
...
-  (cite:[ACPI] Section 5.6)
-* There is at least 1 wake event.
```
andreiw, line 43: Not in the ACPI spec either.

acpi.adoc
```diff
@@ -7,18 +7,9 @@ This section defines mandatory and optional ACPI requirements on top of the ACPI
 
 === ACPI High-Level Requirements
 
-A compliant system must:
-
-* Conform to UEFI specification requirements for ACPI:
```
andreiw, line 12: You've removed a lot, including aspects that are not covered as *requirements* by any other spec. For example, nothing today requires an RHCT, or specific MADT entries (as there isn't a ACPI section for RISC-V).

This change makes the spec useless because it forces the reader to go on a discovery quest to see what the implicit requirements are.

andreiw (commented):
This doesn't fit the spirit of the spec, which is meant to be concise and useful. The BRS spec /points out/ those aspects which are not necessary clear from the (very large) dependent specs, to avoid ambiguity and the need to reference existing implementations just to chart out a course of action.

andreiw (approved):

andreiw (commented):
Only meant to approve this one commit ("remove redundant language")

andreiw (changes requested):
Only meant to approve this one commit ("remove redundant language")

andreiw (commented):
How is this insignificant, when the ACPI section talks about TAD?

about.adoc
```diff
@@ -19,6 +19,5 @@
 | SMBIOS          | System Management BIOS cite:[SMBIOS].
 | SoC             | System on a chip, a combination of processor and supporting chipset logic in single package.
 | System          | A system is the entirety of a computing entity, including all elements in a platform (hardware, firmware) and software (hypervisor, operating system, user applications, user data). A system can be thought of both as a logical construct (e.g. a software stack) or physical construct (e.g. a notebook, a desktop, a server, a network switch, etc).
-| TAD             | Time and Alarm Device cite:[ACPI] Section 9.17.
```
andreiw, line 22: How is this insignificant, when the ACPI section talks about TAD?

recipes.adoc
```diff
@@ -11,34 +11,11 @@ and BRS-B (for "Bespoke").
...
-system-specific customizations - only an implementation of BRS-I
-requirements. *Successfully boot* means basic system configuration,
-sufficient for detecting the need for system-specific drivers and
-loading such drivers.
+BRS-I is expected to be used by general-purpose compute devices such
+as servers, desktops, laptops and other devices with industry
+expectations on silicon vendor, OS and software ecosystem
+interoperatbility.
 
-It is understood that systems will deliver features beyond those covered
```
andreiw, line 27: This specific section was added because it is not covered elsewhere, and does not apply to BRS-B systems at all.

andreiw: The notion of being able to boot "generic OS" images with no ill side effects is one that makes sense for server, PC, laptop and other horizontal markets (e.g. BRS-I).

acpi.adoc
darius-bluespec: ACPI Specification, Relase 6.5, p 290:

> Hardware-Reduced ACPI platforms use GPIO-signaled ACPI Events, or Interrupt-signaled ACPI events. Note that
any ACPI platform may utilize GPIO-signaled and/or Interrupts-signaled ACPI events (in other words, these events
are not limited to Hardware-reduced ACPIvplatforms).

acpi.adoc
darius-bluespec: It is implied by the fact that the system is able to be turned on.

acpi.adoc
darius-bluespec: > This doesn't fit the spirit of the spec, which is meant to be concise and useful. The BRS spec /points out/ those aspects which are not necessary clear from the (very large) dependent specs, to avoid ambiguity and the need to reference existing implementations just to chart out a course of action.

I agree that the spirit of the specification is to be concise and useful and this change does just that.

If it is valuable to point out anything in another specification, that does not belong in the normative BRS Specification.  Repeating and slightly rewording things from elsewhere will create ambiguity and give the reader even more text to wade through in order to understand the specification.

about.adoc
darius-bluespec: TAD is written exactly twice in the entire document.  First, where it is defined inline, and second, in the latter half of **the** **same** **sentence**.  Thus, insignificant.

MADT, PPTT, MCFG, ECAM, RINTC, SPCR, AML, UART, and PCIe are also missing from the glossary, among others. Shall we copy the entire ACPI and UEFI glossary into the BRS?

recipes.adoc
darius-bluespec: It also makes sense for for a BRS-B system to be able to do the same.  It also makes sense for any other BRS recipe system to do likewise.  This is described in the introduction to the BRS Specification.

acpi.adoc
adurbin-rivos: Is it wake from off or suspend? If it's off, yes, it's implied a system can be turned on. If the latter that's a different issue.

That said, Andrei, what are you wanting to obtain through this requirement?

acpi.adoc
```diff
@@ -15,8 +15,7 @@ A compliant system must meet the following requirements:
 
 === ACPI-specific Hardware Requirements
 
-This section lists the BRS-specific hardware requirements imposed by
-ACPI support.
+This section lists the requirements imposed by ACPI support.
```
adurbin-rivos, line 18: Darius, could you pull out this commit into a separate PR so we can merge this one?

darius-bluespec: #38

adurbin-rivos: Thanks. Applied.

acpi.adoc
andreiw: So this sounds like a reference to the section is useful.

acpi.adoc
andreiw: Ok, we can drop the "at least 1 wake event" part.

acpi.adoc
andreiw: "Today" is 6.5. 6.6 isn't out and doesn't have a chance at containing everything we'd like ACPI to be precise on.

The BRS specification is not the Dummy's Guide to ACPI and UEFI, but you would be surprised at the quality of implementations when you try to be least helpful to the very people you expect to rely on for your success. I'm fine with moving things into non-normative, but the references need to exist. You can't expect people to read 1200 pages of the ACPI spec and almost 3000 pages of the UEFI spec (and there are other specs too...), and nail down an implementation without so much as a hint in the right direction.

about.adoc
andreiw: It would be reasonable to cover those in the glossary, yes.

acpi.adoc
darius-bluespec: The BRS Specification requires ACPI version ">= 6.6".  If that version does not exist, or is not expected to exist, the requirement in the BRS Specification needs to change.  It's hard to have meaningful discussion on this if we can't be clear and consistent about which versions of the dependent specifications we are talking about.

I take no position on whether the extraneous information and restatement of requirement in other specifications is useful or whether or not a guidance document is needed. My point is that the BRS Specification is not the place for it.

If the concern you speak of is the excessive length of the ACPI, UEFI, and other specifications, I question whether making specifications longer is the right solution.

recipes.adoc
darius-bluespec: Actually, yes, the goal of interoperability is stated in the very first sentence of the BRS Specification, in the Introduction:

> The Boot and Runtime Services (BRS) specification provides the software requirements for system vendors and Operating System Vendors (OSVs) to **interoperate** with one another by providing expectations for the Operating System (OS) to utilize in acts of device discovery, system management, and other rich operations provided in this specification.

This applies to the BRS Specification as a whole as well as to all recipes.

Just because the BRS-B isn't concerned with **Very Important Commercial Operating Systems®™**, doesn't mean that interoperability is irrelevant.  The BRS-B, and presumably other future recipes, are concerned with interoperability between different sets of implementations and operating systems.

If you feel that the term "bespoke" strongly indicates a lack of interoperablity, then perhaps that receipe needs a different, more appropriate name.

acpi.adoc
andreiw: Whether you keep 6.5 or 6.6 depends on where we expect this spec to land relative to 6.6 getting ratified. I agree that talking about 6.5 is probably not net useful.

If something is not on track to be in 6.6 - it needs to be here. This may include generic guidance info around the mandatory ACPI tables, for example. When stuff lands in the ACPI it can be removed here. Same goes for the UEFI spec. We're talking about the same set of people working on all of these, so it's not unreasonable.

acpi.adoc
andreiw: Btw, I had filed https://mantis.uefi.org/mantis/view.php?id=2377 to "fix" this in the ACPI spec and no one in ASWG provided feedback that 5.6 ACPI Event Programming Model was a section that already covered this.

You could say this proves something about the length/organization of the ACPI spec, sure, but that's not relevant to our activities on the BRS. As far as the BRS goes, it shows the importance of guiding text pointing out ACPI sections to focus on. Does that go in the normative section? It doesn't (since it *exists* in a dependent spec), but it sure should go into the guidance (non-normative) portion of the BRS document.

about.adoc
andreiw: On a second thought, yes TAD appears to be insignificant because it is used and defined in the same section only.
MADT, PPTT, RINTC, SPCR,   fit this pattern too.

ECAM really needs to be defined in line.

andreiw: Proposed separately via rhttps://github.com/riscv-non-isa/riscv-brs/pull/59

andreiw:
https://github.com/riscv-non-isa/riscv-brs/commit/bd4034545755235156a18c2145ef10df9b1da449, https://github.com/riscv-non-isa/riscv-brs/commit/e09c5eae7cf56c55aae6b92dad1f839418f56229, https://github.com/riscv-non-isa/riscv-brs/commit/d534042723d8f54bc63ecdb62bc8517583c78ad1 - these should take care of the everything around the ACPI portions being too verbose.

If there's anything else missing, please rebase and re-propose. Otherwise, please close the pull request.

darius-bluespec:
> If there's anything else missing, please rebase and re-propose.

As you wish.  The branch has been rebased and updated.  All of the other changes have been applied to main in attributed and non-attributed commits.

PR #44, opened 2023-06-09
Add SBI chapter

atishp04:


atishp04:
@andreiw @adurbin-rivos

sbi.adoc
```diff
@@ -1,3 +1,32 @@
...
+* SBI System Reset (SRST)
+
+* SBI Performance Monitoring (PMU)
+
+In addition to this, the following SBI extensions are only mandatory if required
+ISA extension is not available in the platform. 
+
+* SBI TIME 
+** This extension is mandatory only if SSTC ISA extension is not available.
+* SBI IPI
```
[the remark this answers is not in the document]
adurbin-rivos: Oh. I see the 'these two' language below. It may be better to just repeat the requirement for !IMSIC. Or it's ok just grouping them together. I don't have much of an opinion.

sbi.adoc
```diff
@@ -1,3 +1,32 @@
 [[sbi]]
 == SBI Requirements
 
+The supervisor binary interface (SBI) cite: [SBI] specification defines the interface
+between the supervisor mode and the next higher privilege mode. This section
+defines the mandatory SBI version and extensions implemented by the higher
+privilege mode in order to be compatible with this specification.
+
+=== Minimum SBI version
+The SBI implementation must be v1.0 or higher.
```
xypron, line 10: The wording should explicitly relate to the specification version and not to an implementation version, e.g:

The SBI implementation must conform to the SBI v1.0 specification.

xypron: "v1.0 or higher" would allow to conform to a future version even if it contradicts the requirements of v1.0. That is why I think we should relate to a specific version.

sbi.adoc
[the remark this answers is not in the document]
atishp04: Sure. Will do

sbi.adoc
```diff
@@ -1,3 +1,32 @@
...
+The SBI implementation must be v1.0 or higher. 
+
+=== Mandatory SBI extensions
+
+The SBI v1.0 allows optional SBI extensions. However, the following optional
+extensions are mandatory for a compliant system. 
+
+* SBI Hart State Management (HSM)
+
+* SBI System Reset (SRST)
```
andreiw, line 19: Is that truly mandatory? OSes are going to be mandated (by this spec) to always use UEFI RT call to perform a reset or shudown, and conceptually it doesn't matter if the UEFI RT SystemReset call is backed by SRST or something else.

sbi.adoc
```diff
@@ -1,3 +1,32 @@
...
+
+* SBI Performance Monitoring (PMU)
```
andreiw, line 21: Is this really mandatory to booting and general operation? Can this be retrofitted to existing systems?

andreiw (commented):
Just a couple small questions.

sbi.adoc
xypron: Having SRST avoids having to implement system reset drivers in the UEFI firmware. In the ARM world PSCI is also on the TF-A level.

sbi.adoc
atishp04: sgtm.

sbi.adoc
atishp04: If you don't consider "performance monitoring" as a part of general operation, we can ignore that.
Yes, SBI PMU extension works on existing systems with limited capability.
i.e. You can run perf stat on hifive unmatched with SBI PMU for all the hardware hpm counters available.

sbi.adoc
atishp04: We can add more clarification saying UEFI firmware (running in S-mode) must implement the ResetSystem service via SBI SRST extension.

atishp04: I have updated the text. Please let me know if the clarification is sufficient.

atishp04:
We still have one open regarding SBI debug console. I have not added that yet. Let me know if we should make that mandatory as well to avoid specifying an UART type later in the SoC/platform spec

sbi.adoc
atishp04: The below clarification is meant for both suspend and reset.

adurbin-rivos: I think I was confused, and it looked like a new PR was created. I think we're all good. Thanks, Atish.

PR #54, opened 2023-08-02
Add a UEFI-PI reference to the bib file.

andreiw:
Reference a yet-non-existing 1.9 version of the spec.


andreiw:
Trivial fix for an error (warning?) message during build, so merging.

PR #69, opened 2023-08-28
fill out the uefi section

andreiw:


uefi.adoc
```diff
@@ -1,37 +1,52 @@
...
+
+This section defines the BRS-I mandatory and optional UEFI requirements on top of cite:[UEFI], and is optional and recommended for BRS-B. Additional non-normative guidance may be found in the <<uefi-guidance, appendix>>.
+
+A compliant system must:
+
+* Implement a 64-bit UEFI firmware.
+* Meet the 3rd Party UEFI Certificate Authority (CA) requirements on UEFI memory mitigations cite:[MSUefiCaRequirements].
+* Meet the following UEFI memory map requirements:
+** The default memory space attribute must be EFI_MEMORY_WB.
+** Enable address translation.
```
adurbin-rivos, line 14: What is this supposed to mean? Enable paging? When should it turned be off?

uefi.adoc
```diff
@@ -1,37 +1,52 @@
...
+*** Any memory space defined by the UEFI memory map must be identity mapped (virtual address equals physical address), although the attributes of certain regions may not have all read, write and execute attributes or be unmarked for purposes of platform protection.
```
adurbin-rivos, line 15: These items seem to be extracted from ia32 requirements, correct? Shouldn't we feed that directly back into UEFI spec?

uefi.adoc
```diff
@@ -1,37 +1,52 @@
...
+*** The mappings to other regions are undefined and may vary from implementation to implementation.
+*** The highest hart-supported MMU configuration must be used. E.g., Sv39 should not be used on an Sv48-capable chip.
```
adurbin-rivos, line 17: What is the motivation for this? I don't quite understand the motivation for this.

uefi.adoc
adurbin-rivos: should we add more specificity or 'address translation sufficient'?

uefi.adoc
adurbin-rivos: Applies to DXE Boot Services only. Implied motivation/assumption is using for probing purposes.  Claim is that MMU capability is hard to discover.  https://docs.google.com/document/d/1zSw9eaW3tlrdGxkclo9FsLvYxUvhAjRCR8_71EfRJR4/edit

adurbin-rivos: Exists in RHCT.

uefi.adoc
adurbin-rivos: @andreiw says these requirements are in-flight w/ uefi spec proper to be included. uefi calling conventions sections for risc-v. Approval is supposedly this week (sept 11)

PR #70, opened 2023-08-29
Assorted cleanup

andreiw:


acpi.adoc
```diff
@@ -3,31 +3,20 @@
...
-The following additional requirements apply:
-
-* No FACS table.
-* For the ACPI Platform Error Interfaces (APEI) support:
-** There is 1 event for non-fatal error signaling (cite:[ACPI] Section 18.3.2.7.2)
-** There is 1 NMI-equivalent signal for use in fatal errors (cite:[ACPI] Section 18)
+* [[acpi-hw-reduced]]Implement the HW-Reduced ACPI Mode.
+** No FACS table.
+** <<acpi-guidance-hw-reduced, See additional guidance>>.
+* Implement a firmware-first ACPI Platform Error Interface (APEI) model (cite:[ACPI] Section 18.4).
```
adurbin-rivos, line 17: Does this imply this is the only model? Or that something is exposed as opt-in for firmware first.

adurbin-rivos (approved):

acpi.adoc
andreiw: Are you referring to the APEI portion? I just moved the text, but we can take the APEI requirements out for now if we need to dig into the details here (e.g. my personal opinion is that for BRS-I we mandate APEI, but there may be more work on the APEI side - but that's not really a BRS task by itself)

adurbin-rivos: Ya, I was. I know you were moving things, but I did want to ask this question about the implications. I was mainly talking about 'firrmware-first APEI' as the only option. If that's the case, that's a discussion to be had.  I can put it on the agenda for the next meeting.

andreiw: Ok, let's put that on the agenda. I raised https://github.com/riscv-non-isa/riscv-brs/issues/71 so we don't lose track of it. I'll merge this pull request meanwhile.

PR #72, opened 2023-09-07
Add a new recipe to support non-UEFI firmware (coreboot).

weidongwd:
We want to add a new recipe to support non-UEFI firmware (coreboot), and this recipe doesn't affect the previous recipes.

### DIFF
--- a/non-normative/acpi.adoc
+++ b/non-normative/acpi.adoc
@@ -1,7 +1,8 @@
 ACPI information is structured as tables with the address of the root of these
 tables known as Root System Description Table (RSDP) passed to the OS
-via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware.
-The Operating System uses this address to locate all other ACPI tables.
+via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware. In
+addition, RSDP can be obtained in other ways, such as DT. The Operating System
+uses this address to locate all other ACPI tables.
 
 Certain implementations may make use of the RISC-V Functional Fixed hardware Specification cite:[FFH].
 
--- a/non-normative/recipes.adoc
+++ b/non-normative/recipes.adoc
@@ -30,3 +30,18 @@ configuration utilities. The actual meaning of the default (compatible to
 BRS-I) configuration would be highly specific to the vendor. In this example,
 it could capping all harts at the same performance level, or it could mean
 disabling efficiency harts.
+
+[[recipe-brs-m-guidance]]
+===== BRS-M Recipe Guidance
+
+In contrast to BRS-I and BRS-B, BRS-M has no implementation of UEFI related
+services. Non-UEFI firmware, such as *coreboot*, can also supports boot well.
+Although non-UEFI firmware does not support UEFI interface, it can also
+provide ACPI, SMBIOS and SBI services. ACPI and SMBIOS can be passed through
+DT.
+
+The introduction of coreboot can refer to https://www.coreboot.org/[coreboot].
+This solution is the startup mode for coreboot plus payload. coreboot does the
+basic and necessary hardware initialization, where ACPI and SMBIOS tables are
+created, and then loads and executes payload, which completes the boot and
+additional functions. linuxboot is a common payload.
--- a/recipes.adoc
+++ b/recipes.adoc
@@ -6,8 +6,8 @@ requirements that hardware, firmware, and software providers can
 implement to increase the likelihood that software written to the
 recipe will run predictably on all conforming devices.
 
-The BRS specification defines two recipes: BRS-I (for "Interoperable")
-and BRS-B (for "Bespoke").
+The BRS specification defines three recipes: BRS-I (for "Interoperable"),
+BRS-B (for "Bespoke") and BRS-M (for "Minimal").
 
 === BRS-I Recipe
 
@@ -65,4 +65,21 @@ custom distributions.
 |===
 | Profile | UEFI | ACPI | DT | SBI | SMBIOS
 | >= RVA20 | EBBR cite:[EBBR] | optional, >= 6.6 | optional, >= v0.3 | >= 2.0 | optional, >= 3.6.0
+|===
+
+=== BRS-M Recipe
+
+BRS-M is used for cases that do not require UEFI. BRS-M is the
+smallest recipe. This recipe does not require UEFI boot and runtime
+service implementations. In some boot scenarios, non-UEFI firmware
+can also support boot, they only focus on the boot process and support
+the necessary interaction interfaces.
+<<recipe-brs-m-guidance, See additional guidance>>.
+
+.BRS-M Recipe Overview
+[width=100%]
+[%header, cols="10,20,20,10,20"]
+|===
+| Profile | ACPI | DT | SBI | SMBIOS
+| >= RVA20 | optional, >= 6.6 | optional, >= v0.3 | >= 2.0 | optional, >= 3.6.0
 |===
--- a/about.adoc
+++ b/about.adoc
@@ -10,6 +10,7 @@
 | BRS             | RISC-V Boot and Runtime Services. This document.
 | BRS-I           | Boot and Runtime Services Recipe targeting interoperation across different software suppliers.
 | BRS-B           | Boot and Runtime Services Recipe using a bespoke solution.
+| BRS-M           | Boot and Runtime Services Recipe using a minimal solution.
 | DT              | DeviceTree cite:[DT].
 | EBBR            | Embedded Base Boot Requirements Specification cite:[EBBR].
 | OSV             | Operating System Vendor.
--- a/intro.adoc
+++ b/intro.adoc
@@ -11,7 +11,7 @@ It is expected that the Boot and Runtime Services specification will periodicall
 
 === Approach to Solutions
 
-The Boot and Runtime Services specification focuses on two solutions in the form of what is deemed a recipe. Each recipe contains the requirements needed to fulfill each solution. The requirements of each recipe will be marked accordingly with an unique identifier. The recipe names are named BRS-I and BRS-B, Interoperable and Bespoke, respectively.
+The Boot and Runtime Services specification focuses on three solutions in the form of what is deemed a recipe. Each recipe contains the requirements needed to fulfill each solution. The requirements of each recipe will be marked accordingly with an unique identifier. The recipe names are named BRS-I, BRS-B and BRS-M, Interoperable and Bespoke, respectively.
 
 === Testing and Conformance
 


weidongwd:
@atishp04 :Could you please review this PR？

cuiyunhui-1:
@andreiw @xypron @atishp04 @darius-bluespec 

Please review the patch, Thanks!

non-normative/recipes.adoc
```diff
@@ -30,3 +30,18 @@ configuration utilities. The actual meaning of the default (compatible to
 BRS-I) configuration would be highly specific to the vendor. In this example,
 it could capping all harts at the same performance level, or it could mean
 disabling efficiency harts.
+
+[[recipe-brs-m-guidance]]
+===== BRS-M Recipe Guidance
+
+In contrast to BRS-I and BRS-B, BRS-M has no implementation of UEFI related
+services. Non-UEFI firmware, such as *coreboot*, can also supports boot well.
```
adurbin-rivos, line 38: put coreboot in the bibtex and reference it from here?

adurbin-rivos:
would someone be willing to come to BRS meeting this coming Monday and present to the group on this new recipe?

andreiw:
Would you be willing to attend the next BRS call and share your proposal so we can have a group discussion?

Personal opinions below.

From a technical standpoint, there are a million ways anything could be done, but the overarching question for a specification remains - what distinct benefit will a standard bring to the a product, a market, a producer (vendor, manufacturer) or a consumer (whether an end consumer, or an OEM or an OSV). For example, the BRS-I and BRS-B are distinct in that the first focuses on PC-like systems (maximum compat, "horizontal integration" with distinct SW and solution/platform vendors), while the second focuses on embedded/vertical markets, and thus sheds many requirements. When folks read the BRS spec, they will know exactly which recipe to follow.

Your proposal introduces an alternate hand-off interface between firmware and the OS. This recipe, like BRS-I and BRS-B can sustain either DT or ACPI descriptions. Unlike either BRS-I or BRS-B, it does not standardize on BS/RT services as the boot loader interface. From the current text, it's not clear what the benefit of this is, so my recommendation is to at least spend the time and extend the intro paragraph so the area of applicability for such a BRS-M is obvious.

In a BRS-I and BRS-B world, a compliant implementation could be based on coreboot (the "UEFI" personality is a payload, coreboot takes place of existing PI components).

In a BRS-I and BRS-B world, a compliant implementation could be based on linuxboot. This is LinuxBoot booting an OS that expects UEFI interfaces, i.e. the use of a Linux or other rich OS kernel to access I/O, instead of having to reimplement things in U-Boot or Tiano, does not preclude UEFI or a subset thereof as the interface to the booting OS loader. For some context, see https://uefi.org/sites/default/files/resources/LinuxBoot_FINAL_0.pdf and https://developer.arm.com/Architectures/Arm%20SystemReady%20LS (and LBBR https://developer.arm.com/documentation/den0044/latest)

Now there is LinuxBoot adoption in the cloud. Perhaps this is your use case, in which case there's no denying this is an important area to define for RISC-V, but it doesn't have to imply a complete lack of interoperability as far as the hand-off interface goes. Take a look at https://developer.arm.com/documentation/den0044/latest for the LBBR requirements in Arm SystemReady LS, specifically section 10.1 "LBBR-v1 requirements". In short, this defines an even smaller set of UEFI requirements that LinuxBoot is capable of implementing. At the very least, you get to avoid the unnecessary additional methods of describing ACPI via DT, and adopting ACPI sends a clear message around the fitness of the hardware for the use case.

In general, many conflate UEFI implementations (such as the opensource TianoCore, TianoCore-derived closed source implementations, entirely closed source implementations like Insyde, and the opensource U-Boot) with the UEFI specification. There's no "UEFI firmware", it's a series of requirements on functionality. LinuxBoot can easily be "UEFI firmware" itself (as per LBBR), constraining itself to just the market-required functionality. Over the course of the last 10 years the U-Boot folks realized the real benefits of UEFI boot services over their own proprietary (in the sense of being bespoke or their own) bootloader interfaces, esp. as this relates to booting existing distros - and this is how EBBR came to be (present in SystemReady IR or the RISC-V BRS-B). Similarly, the Arm SystemReady LR (LinuxBoot) have reached a similar conclusion.

non-normative/recipes.adoc
```diff
@@ -30,3 +30,18 @@ configuration utilities. The actual meaning of the default (compatible to
...
+Although non-UEFI firmware does not support UEFI interface, it can also
+provide ACPI, SMBIOS and SBI services. ACPI and SMBIOS can be passed through
```
andreiw, line 40: Because this is a spec, you need to say how that would be done (or reference another spec). Again, as per my comment in https://github.com/riscv-non-isa/riscv-brs/pull/72, if you want to go this route to enable cloud-scale LinuxBoot, follow the Arm SystemReady LR appoach with their LBBR requirements and use an even more constrained subset of UEFI (compared to EBBR). You still get the consistent hand-off interface, and you don't need to hack up booting OSes to understand your new hand-off structures.

https://developer.arm.com/documentation/den0044/latest, see 10.1 LBBR-v1 requirements

cuiyunhui-1: In fact, UEFI is not supported in coreboot and will not be supported in the future.

See https://lore.kernel.org/lkml/ZKgLKvBoWKSxzm6r@sunil-laptop/T/
As Coreboot maintainer Ron said: 

part1:
ACPI was not originally part of UEFI. ACPI works just fine on Chromebooks, and has for 12 years, and on coreboot since 2006, without UEFI. I've integrated support for ACPI into several code bases, including Plan 9 on non-UEFI systems.

I am unable to understand the claim that ACPI on non-UEFI RISC-V is a bad idea. Clearly, I am not alone. 

part2:
Glad to see we're not going to revisit UEFI arguments. I've been doing this dance for 25 years, and I've seen the suggestion that adding UEFI plumbing to coreboot is easier, and we've actually tried that out several times in the last 16 years, and ... it's really not worked well for many reasons, some of them involving the many ambiguities in the spec that are answered in proprietary Intel code (which Intel engineers have pointed out to me more than once).  

And that's why since 2012 chromebooks have run coreboot, without any UEFI interfaces of any kind, even though the very first chromebooks were based on UEFI. Whether or not UEFI is the standard, it is hardly universal.

I'm trying hard not to get sucked into this argument, but it is, at least, impolite, to accuse us of engaging in hand-wavery. There's probably more UEFI expertise in coreboot and oreboot than most places, and that's precisely why we want a non-UEFI option. Familiarity with UEFI did not breed respect.

So can we get back to the business at hand? bytedance may be the company proposing this change, but they are certainly not the only company interested in seeing it happen. A project I'm involved in needs this work to go through.

xypron: On x86 Coreboot can launch U-Boot (https://u-boot.readthedocs.io/en/latest/board/coreboot/index.html). The same could be implemented for RISC-V. U-Boot can provides the UEFI API. Coreboot and UEFI are just different firmware levels and not contradictory.

For UEFI we have a standardized way to pass ACPI and SMBIOS tables to Linux. According to [RISC-V Kernel Boot Requirements and Constraints](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/riscv/boot.rst?h=v6.6-rc1) such an interface does not exist for the legacy entry point.

Referring to ACPI makes only sense if such an interface is defined or linked to.

non-normative/recipes.adoc
```diff
@@ -30,3 +30,18 @@ configuration utilities. The actual meaning of the default (compatible to
...
+DT.
+
+The introduction of coreboot can refer to https://www.coreboot.org/[coreboot].
+This solution is the startup mode for coreboot plus payload. coreboot does the
```
xypron, line 44: This paragraph sounds more like a commit message than anything that would go into a standard.

It remains unclear why Coreboot is mentioned here. Is Coreboot a requirement of BRS-M?

If yes, say so. If not, define interfaces without relating to Coreboot.

non-normative/recipes.adoc
xypron: Why do you relate to Coreboot here? Isn't any SBI implementation enough to boot an OS (e.g. Linux).

non-normative/acpi.adoc
```diff
@@ -1,7 +1,8 @@
 ACPI information is structured as tables with the address of the root of these
 tables known as Root System Description Table (RSDP) passed to the OS
-via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware.
-The Operating System uses this address to locate all other ACPI tables.
+via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware. In
+addition, RSDP can be obtained in other ways, such as DT. The Operating System
+uses this address to locate all other ACPI tables.
```
xypron, line 5: This is very vague. Where are these other ways defined?

recipes.adoc
```diff
@@ -66,3 +66,20 @@ custom distributions.
...
 |===
+
+=== BRS-M Recipe
+
+BRS-M is used for cases that do not require UEFI. BRS-M is the
+smallest recipe. This recipe does not require UEFI boot and runtime
+service implementations. In some boot scenarios, non-UEFI firmware
+can also support boot, they only focus on the boot process and support
+the necessary interaction interfaces.
+<<recipe-brs-m-guidance, See additional guidance>>.
```
xypron, line 77: In this merge request I only find what BRS-M does not offer but not any definition of what it should provide. This is insufficient to define a recipe.

xypron (changes requested):

non-normative/recipes.adoc
weidongwd: Because UEFI is not the only firmware interface, UEFI is not used in all application scenarios. So we want to support more firmware solutions than UEFI.

non-normative/acpi.adoc
weidongwd: We already have a whole set of schemes based on device tree, and have completed the preliminary implementation and verification. This part of the solution is also being prepared upstream.

recipes.adoc
weidongwd: What do you suggest should be added, where?

non-normative/recipes.adoc
weidongwd: We believe that the firmware ecosystem is developed and should not be limited to UEFI alone. In this way, it can also adapt to more scenarios.
From the BRS spec, there is no direct description of UEFI and ACPI binding. For example, coreboot can easily support ACPI/SMBIOS, but if you add EFI stub, it adds complexity and no better benefits.

non-normative/recipes.adoc
adurbin-rivos: My suggestion was about citing the external references. Since coreboot is newly introduced it should be cited.

ConchuOD:
@vlsunil It may be worth re-iterating some of what you said on LKML here for the new audience.

vlsunil:
Sure Conor. Below is my opinion which I have communicated multiple times.

The proposal tries to remove UEFI requirement for ACPI and SMBIOS support on RISC-V. However,  ACPI and SMBIOS specs clearly document how their entry points can be found by the OSs. Refer to section 5.2.2 of SMBIOS spec (https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.4.0.pdf) and Section 5.2.5.1/5.2.5.2 of ACPI spec (https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#finding-the-rsdp-on-ia-pc-systems).

Clearly, RISC-V can not support legacy x86 way and the only other way ACPI/SMBIOS specs define is to use UEFI. As per Dong Wei, even the way coreboot supports ACPI without UEFI on x86 is a kludge  (REF: https://lore.kernel.org/lkml/ED8788D4-82D9-4B60-88F5-0FA5F00E4A36@arm.com/).

So my recommendation is, please work with ACPI/DMTF WG to define a new way of finding the ACPI/SMBIOS entry points (and other things like memory map) in architecture neutral way on non-UEFI platforms.

non-normative/recipes.adoc
andreiw: To shorten my comments: prove that you need a different way of approaching things than the Arm LBBR, which seems to tick all the checkboxes you desire (no full UEFI, suitable for Linux-based or alternative firmware), and is an industry accepted approach. You can't just propose a completely ad-hoc mechanism and not expect pushback towards something that has wider acceptance.

andreiw: Btw, would you be so kind and subscribe to tech-brs@lists.riscv.org? We could have discussed this in the meeting today and  listened to the proposal, but it wasn't brought up as an agenda item (nor did you get back to us on our request to have this presented).

My recommendation is to quench the emotions a bit, avoid hyperbole and try to understand each other. This is where, again, a presentation to the BRS team is going to be helpful. The overall approach to the BRS spec is to adopt existing functional solutions and stick to industry approaches instead of inventing new ones. This is why BRS largely copies the Arm BBR approach, and why we take an aggressive policy of not duplicating requirements seen in upstream specs (but *do* provide a non-binding guidance document to help people navigate the specs). Again, to reiterate - I personally get your reasons, but there's already an industry approach (LBBR) that I think you should investigate first before rolling your own, and have concrete feedback on what exactly doesn't work about it for you (because it sure works for other hyprescalers). I pointed you to the LBBR section, it's pretty short.

non-normative/acpi.adoc
andreiw: Hate to be a pedant, but the point of a spec is exactly "to avoid a whole set of schemes", and you should have the mechanism well identified and described in the spec as part of your pull request, not as a promise. As it stands if someone were to implement your pull request today, they couldn't do it.

non-normative/recipes.adoc
andreiw: Btw, as part of the BRS/PRS work we have identified gaps in UEFI/ACPI and have worked to address these. So, if you have concrete examples of something that is "the many ambiguities in the spec that are answered in proprietary Intel code", we'd be happy to take that flag up in the USWG (or you can raise these yourself, either via UEFI Forum USWG or the OSS Code First Process @ https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-First-Process).

andreiw:
> > it does not standardize on BS/RT services as the boot loader interface. From the current text, it's not clear what the benefit of this is
> 
> The benefit is that, as soon as the boot process is gone, the boot code is gone. One less actor that might possibly create a mess. Of course, that needs to be accompanied by getting rid of SBI for full effect, but for security conscious folks, doing both of these is an interesting proposition simply because it minimizes the attack surface in privileged environments.

Take a look at LBBR (10.1 LBBR-v1 requirements in https://developer.arm.com/documentation/den0044/latest/), and do let me know if this addresses your immediate requirements (around not implementing BS and RT).

non-normative/recipes.adoc
adurbin-rivos: I mistook xypron's comment as a response to my comment.  That seems to not be the case.

xypron:
Am 18. September 2023 21:13:31 MESZ schrieb Patrick Georgi ***@***.***>:
>LBBR-v1 10.1: "With this approach, LinuxBoot kernel exposes the required interfaces (UEFI Runtime, ACPI, and SMBIOS) to be used by the final runtime Linux kernel"
>
>That's already one component too many: "UEFI Runtime".
>
>"EFI_SYSTEM_TABLE and EFI_RUNTIME_SERVICES are required for LBBR-v1 systems. All UEFI System Table Function pointers (including the UEFI Runtime Services) must be available and cannot be NULL pointers."
>
>Same issue.
>
>With everything else in that list in 10.1.3.2 being "optional", it makes me wonder why `SetVirtualAddressMap` and `ConvertPointer` are still required.

SetVirtualAddressMap informs the EFI runtime about the address translations that will be set up by the OS. If you don't implement any of the optional services, you will still have to update the system table accordingly.

ConvertPointer is only used internally by your runtime implementation in the LBBR case. If you don't use it, returning EFI_UNSUPPORTED is good enough.

>
>(By the way, LBBR-v1 says "This document references the following specification and versions:
>UEFI 2.9 [25] Published March 2021, includes the AArch64 bindings". That version of UEFI is improperly licensed for an independent implementation.)

EDK II is BSD licensed. So anything implemented there can be reused accordingly.

>
>It's interesting that UEFI proponents recently started bringing up "But UEFI-the-spec is not UEFI-the-implementation" to deflect from all its faults. Ever considered that a spec can't be all that good if in ~20 years nobody managed to implement it reasonably?
>

Your recipe suggestion needs to become as specific as the UEFI specification to be implementable.

Best regards

Heinrich

weidongwd:
> > > That's already one component too many: "UEFI Runtime".
> 
> > SetVirtualAddressMap informs the EFI runtime about the address translations that will be set up by the OS. If you don't implement any of the optional services, you will still have to update the system table accordingly.
> 
> I have to setup a bunch of tables (e.g. memory mapping), but I don't have to tell any runtime services about it, **if there are no runtime services**. That's the whole point!
> 
> > > (By the way, LBBR-v1 says "This document references the following specification and versions:
> > > UEFI 2.9 [25] Published March 2021, includes the AArch64 bindings". That version of UEFI is improperly licensed for an independent implementation.)
> 
> > EDK II is BSD licensed. So anything implemented there can be reused accordingly.
> 
> As folks recently started to love pointing out: UEFI is the specification, not the implementation.
> 
> Let me try as well: UEFI-the-spec has a very different license than BSD-l, that attaches a whole bunch of strings to it. To quote https://uefi.org/specifications:
> 
> > By downloading any of the UEFI Specifications, you acknowledge that no license, express or implied, is granted to you to distribute, additionally reproduce, implement or otherwise use for any purpose (other than to read only) the UEFI Specifications

Yes, we don't think it's necessary to follow UEFI if you only pass the necessary system information without supporting UEFI BS/RT services.

non-normative/acpi.adoc
weidongwd: We are now promoting and improving this recipe into a mechanism. As for DT-based passing mechanisms, we are also pushing DT spec to support ACPI/SMBIOS pointer passing.

ConchuOD: > We are now promoting and improving this recipe into a mechanism. As for DT-based passing mechanisms, we are also pushing DT spec to support ACPI/SMBIOS pointer passing.

It is probably worth noting that I NAKed the DT side of this, as it violated "UEFI firmware is mandatory to support ACPI" in https://github.com/riscv-non-isa/riscv-acpi/blob/master/riscv-acpi-guidance.adoc. Not gonna be changing my mind on that unless the specs get changed first.

cuiyunhui-1: I've already said it: this spec is obsolete and no one maintains it, and this guidance was only written by sunil, which was also discussed with sunil @vlsunil 

How about deleting this guidance? So as not to cause misunderstanding.

vlsunil: That's right. The guidance document was created long back when platform spec was being drafted. This was never intended to be a spec and neither ratified/frozen. Also, now things have changed to BRS spec which has more official version of the similar requirements. The guidance document  has old things like only AIA will be supported with ACPI etc which is no longer the case. So, there is no value in keeping this any more.  I will be deleting this guidance document soon in favor of BRS spec.

ConchuOD: FWIW, that doesn't change my position. I don't care where the rules are defined, but unless they clearly say that UEFI is not required to use ACPI the NAK stands.

NillatBD: ACPI ans UEFI were two seperated spec before, both are interfaces between FW and OS, and ACPI was moved to UEFI org since ARM platform need to support both UEFI and ACPI; in the ACPI soec 5.2.5, it defined RSDP and also list two solutions how to find the RSDP on IA-PC and UEFI edabled systems, they are for reference ony. 
With current solution, it's on non-UEFI enabled system, and there is NOT EBDA on non-x86 platform, so I think this is great solution to find the RSDP from DTS.

vlsunil: Sorry, I don't agree that the mechanisms mentioned in ACPI spec to find the RSDP is only for reference. Because that will imply that the platforms are free to pass RSDP in any arbitrary manner and OSPMs can find them which is not true. The FW and OSs need a known contract and that's why those two mechanisms are documented in the ACPI spec. Same applies to SMBIOS spec also.

I fail to understand why are you reluctant to propose change to ACPI/SMBIOS specs. They are the original place where UEFI is mandated. The proposal has nothing really which applies to only RISC-V. So,if you want to remove UEFI requirement, ACPI/SMBIOS specs should be updated with new mechanism to discover ACPI/SMBIOS.

NillatBD: In Chrome book, how could firmware pass the ACPI and SMBIOS points to Chrome OS? and could you share with us which spec has cover it?

vlsunil: Sorry I don't know about chromeOS. I am keen to learn that it doesn't use either of the ways defined by the ACPI spec (EBDA or UEFI). Could you find the answer and provide more information of the mechanism?

adurbin-rivos: From my working on Chrome OS only x86 systems utilized ACPI. I'm assuming that's still true. On x86 the kernel goes and searches in low memory for the expected entries.

https://review.coreboot.org/plugins/gitiles/coreboot/+/refs/heads/main/src/arch/x86/tables.c

rminnich:
There are a number of non-UEFI firmware projects out there: 
tinygo
tamago
oreboot
coreboot
to name 4. 
Also, the original u-BMC ran less than 1000 assembly instructions on the AST2500 (to get DRAM up) and directly loaded and booted a flash-based Linux. That's a fifth.

Historically, ARM has been incredibly accommodating of those projects. Dong Wei in particular has been very helpful with LinuxBoot and the LBBR. An original goal of LBBR, still of interest, is direct loading Linux from TF-A, bypassing UEFI entirely, so we can remove UEFI.

I'd like to see the RISC-V community be at least as accommodating as ARM. To put it another way, the software side of RISC-V should be as open as the hardware side to different uses and extensions.

non-normative/recipes.adoc
rminnich: "Because this is a spec, you need to say how that would be done (or reference another spec). "
Generally, in a spec, you set requirements -- you need to say that it must be done -- but not how. Perhaps I am misunderstanding your wording.
In any event, the wording is "it can also provide ACPI, SMBIOS and SBI services." -- operative word, "can"

For a platform I am working on, I don't want the mess that comes with ACPI, SMBIOS, and I am trying to have SBI do as little as possible. ACPI in particular, due to ACPI ML, is a serious problem to be dealt with by removal.

Further, the coreboot port, which had an SBI that predates OpenSBI, showed how it could be done: it conformed to the interface defined in RISC-V standards. 

I'd be just as happy with wording to this effect: "platforms must support minimal SBI standards, as defined in the RISC-V privelege specification; ACPI and SMBIOS are optional." or some such and leave it that. Will that do?

weidongwd:
Now there are many controversies, let me summarize briefly:

1. Is UEFI necessary for RISC-V
Some people think that UEFI is a standard interface that we should all use to interact with system information. 
First of all, I appreciate UEFI's contribution. However, in some cases, supporting UEFI will not help the firmware implementation, and even cause some confusion. I agree that RISC-V is an open architecture, and the firmware implementation should be open as well, which will help the whole ecosystem. In other words, in the firmware implementation, I want to have both UEFI firmware and non-UEFI firmware support so that users can choose the right firmware solution for their needs.

2. How to pass ACPI/SMBIOS
At present, the ACPI/SMBIOS specification defines two delivery methods, but I think this is not a limit to the delivery method, as long as the function can be implemented according to the specifications. Of course, if people still have concerns about the specifications, we will also try to change these specifications to more explicitly support more delivery methods.
All in all, I think adding a more convenient ACPI/SMBIOS delivery method would be good for RISC-V in general.

adurbin-rivos:
There is a BRS meeting Monday Sept 25 at 9am PT. Will be willing to come and discuss the topics at hand?

weidongwd:
> There is a BRS meeting Monday Sept 25 at 9am PT. Will be willing to come and discuss the topics at hand?

Ok, we will attend the meeting on time.

non-normative/acpi.adoc
adurbin-rivos: Ron, on the "the forthcoming UPL standard will define a new one that is DT-based": is that about finding the ACPI root table? And would the kernel be changed to leverage that path as well?

ardbiesheuvel: The problem with using DT to discover ACPI is the fact that DT and ACPI are both hardware descriptions, and are usually mutually exclusive. Using both at the same time forces the OS to reason about whether or not they are mutually consistent, and about which description should take precedence when they are not.

xypron:
On 9/25/23 23:39, Ard Biesheuvel wrote:
> ***@***.**** commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In non-normative/acpi.adoc 
> <https://github.com/riscv-non-isa/riscv-brs/pull/72#discussion_r1336414034>:
> 
>> @@ -1,7 +1,8 @@
>   ACPI information is structured as tables with the address of the root of these
>   tables known as Root System Description Table (RSDP) passed to the OS
> -via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware.
> -The Operating System uses this address to locate all other ACPI tables.
> +via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware. In
> +addition, RSDP can be obtained in other ways, such as DT. The Operating System
> +uses this address to locate all other ACPI tables.
> 
> The problem with using DT to discover ACPI is the fact that DT and ACPI 
> are both hardware descriptions, and are usually mutually exclusive. 
> Using both at the same time forces the OS to reason about whether or not 
> they are mutually consistent, and about which description should take 
> precedence when they are not.

This is why the EBBR specification explicitly allows only one of ACPI 
and device-tree to be used.

On a technical level the legacy entry point of the Linux kernel already 
uses a device-tree as described in Documentation/arm/uefi.rst to 
transfer the system table from the EFI stub to the kernel as property 
'linux,uefi-system-table' of the /chosen node.

This pointer to the system-table is used to retrieve all EFI 
configuration tables including ACPI and SMBIOS.

Currently this seems to be the only way to transfer an ACPI table to the 
legacy entry point of Linux.

If additionally a EFI_RT_PROPERTIES_TABLE with value 0 is provided, 
Linux will not try to invoke any runtime services.

This discovery of configuration tables is enabled on RISC-V via 
CONFIG_EFI_PARAMS_FROM_FDT.

Hence technically it is already feasible to implement the suggested recipe.

Best regards

Heinrich

non-normative/acpi.adoc
cuiyunhui-1: Can the OS scan UEFI first? If it is not obtained, it can be obtained in the same way as DT.

cuiyunhui-1: Currently in our FFI solution, the RSDP obtained through DT is filled into the acpi_arch_get_root_pointer() interface. The current logic is to obtain it through acpi_arch_get_root_pointer() first. If it cannot be obtained, go to the EFI entrance.

See drivers/acpi/osl.c

pa = acpi_arch_get_root_pointer();
if (pa)
return pa;

if (efi_enabled(EFI_CONFIG_TABLES)) {
if (efi.acpi20 != EFI_INVALID_TABLE_ADDR)
return efi.acpi20;
if (efi.acpi != EFI_INVALID_TABLE_ADDR)
return efi.acpi;
....

cuiyunhui-1: The relationship between ACPI and DTS is compatible in the current kernel, so there is no problem.

Regarding DT transmitting ACPI,  only need to fill in RSDP. ACPI is an independent framework, and EFI and DT are just two different entry interfaces.

weidongwd:
> On 9/25/23 23:39, Ard Biesheuvel wrote: ***@***.**** commented on this pull request. ------------------------------------------------------------------------ In non-normative/acpi.adoc <[#72 (comment)](https://github.com/riscv-non-isa/riscv-brs/pull/72#discussion_r1336414034)>: > @@ -1,7 +1,8 @@ ACPI information is structured as tables with the address of the root of these tables known as Root System Description Table (RSDP) passed to the OS -via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware. -The Operating System uses this address to locate all other ACPI tables. +via a EFI_ACPI_20_TABLE_GUID configuration table in the UEFI firmware. In +addition, RSDP can be obtained in other ways, such as DT. The Operating System +uses this address to locate all other ACPI tables. The problem with using DT to discover ACPI is the fact that DT and ACPI are both hardware descriptions, and are usually mutually exclusive. Using both at the same time forces the OS to reason about whether or not they are mutually consistent, and about which description should take precedence when they are not.
> This is why the EBBR specification explicitly allows only one of ACPI and device-tree to be used. On a technical level the legacy entry point of the Linux kernel already uses a device-tree as described in Documentation/arm/uefi.rst to transfer the system table from the EFI stub to the kernel as property 'linux,uefi-system-table' of the /chosen node. This pointer to the system-table is used to retrieve all EFI configuration tables including ACPI and SMBIOS. Currently this seems to be the only way to transfer an ACPI table to the legacy entry point of Linux. If additionally a EFI_RT_PROPERTIES_TABLE with value 0 is provided, Linux will not try to invoke any runtime services. This discovery of configuration tables is enabled on RISC-V via CONFIG_EFI_PARAMS_FROM_FDT. Hence technically it is already feasible to implement the suggested recipe. Best regards Heinrich

But in essence, ACPI and UEFI are interactive interfaces, and there is no functional dependency. Moreover, DT and ACPI are not explicitly mutually exclusive. DT generally passes system information and configuration through node, so if a configuration node is added to pass ACPI, there should be no problem. Another way of thinking about it is that it just adds a passing interface, not a function.

rminnich:
At the meeting yesterday we agreed that this could be tidied up by removing coreboot references, and it should be possible to get it in.

non-normative/acpi.adoc
adurbin-rivos: This was also discussed that there needs to be alternatives documented in specs for finding ACPI tables.  I indicated that the PRS TG in RVI is the path to achieve that for updating ACPI specs.

Likewise, I also indicated that the kernel maintainers need to be on board as that was where a lot of the disagreement exists. These two things are interrelated which is why the dependencies need to be called out and completed.

weidongwd: Yes, we can merge this PR into BRS spec first.
Regarding ACPI, we are also pushing forward the modification of RSDP search method. At the same time, we are also promoting and discussing linux kernel PR with the community.

non-normative/recipes.adoc
rminnich: can we reduce line 37-47 to this?
BRS-M is intended to support software that has has reduced requirements for firmware, or that has alternative ways to access tables such as ACPI and SMBIOS. Some RISC-V-based systems, such as OpenTitan (see: opentitan.org), do not use tables such as ACPI or SMBIOS at all. In other cases, these tables are provided by new standards such as the Universal Payload Specification (https://github.com/UniversalPayload/spec). Finally, a kernel comtained in firmware might also have the tables comipiled-in or as part of its initial ram file system (initramfs in Linux). 

How does that sound? I'm trying to describe real world scenarios I have used.

recipes.adoc
rminnich: The reason it does not define these things is that they are provided elsewhere / not needed. You can make cake without flour, and people still call it a cake. You can make bread without yeast, and it is flatbread. You can make Linux without block device support, and it is still a kernel. Similarly, BRS-M is a reduced model, but it is still a recipe.

xypron:
On 27.09.23 23:04, ron minnich wrote:
> ***@***.**** commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In non-normative/recipes.adoc
> <https://github.com/riscv-non-isa/riscv-brs/pull/72#discussion_r1339209631>:
>
>> @@ -30,3 +30,18 @@ configuration utilities. The actual meaning of the default (compatible to
>   BRS-I) configuration would be highly specific to the vendor. In this example,
>   it could capping all harts at the same performance level, or it could mean
>   disabling efficiency harts.
> +
> +[[recipe-brs-m-guidance]]
> +===== BRS-M Recipe Guidance
> +
> +In contrast to BRS-I and BRS-B, BRS-M has no implementation of UEFI related
> +services. Non-UEFI firmware, such as *coreboot*, can also supports boot well.
> +Although non-UEFI firmware does not support UEFI interface, it can also
> +provide ACPI, SMBIOS and SBI services. ACPI and SMBIOS can be passed through
> +DT.
> +
> +The introduction of coreboot can refer to https://www.coreboot.org/[coreboot].
> +This solution is the startup mode for coreboot plus payload. coreboot does the
>
> can we reduce line 37-47 to this?
> BRS-M is intended to support software that has has reduced requirements
> for firmware, or that has alternative ways to access tables such as ACPI
> and SMBIOS. Some RISC-V-based systems, such as OpenTitan (see:
> opentitan.org), do not use tables such as ACPI or SMBIOS at all. In
> other cases, these tables are provided by new standards such as the
> Universal Payload Specification
> (https://github.com/UniversalPayload/spec
> <https://github.com/UniversalPayload/spec>). Finally, a kernel comtained
> in firmware might also have the tables comipiled-in or as part of its
> initial ram file system (initramfs in Linux).
>
> How does that sound? I'm trying to describe real world scenarios I have
> used.

A recipe only describing what it not comprised is kind of useless.

We need a description what is in the recipe. Then people taking a board
implementing the recipe can build on these guarantees.

>
> ------------------------------------------------------------------------
>
> In recipes.adoc
> <https://github.com/riscv-non-isa/riscv-brs/pull/72#discussion_r1339214010>:
>
>> @@ -66,3 +66,20 @@ custom distributions.
>   | Profile | UEFI | ACPI | DT | SBI | SMBIOS
>   | >= RVA20 | EBBR cite:[EBBR] | optional, >= 6.6 | optional, >= v0.3 | >= 2.0 | optional, >= 3.6.0
>   |===
> +
> +=== BRS-M Recipe
> +
> +BRS-M is used for cases that do not require UEFI. BRS-M is the
> +smallest recipe. This recipe does not require UEFI boot and runtime
> +service implementations. In some boot scenarios, non-UEFI firmware
> +can also support boot, they only focus on the boot process and support
> +the necessary interaction interfaces.
> +<<recipe-brs-m-guidance, See additional guidance>>.
>
> The reason it does not define these things is that they are provided
> elsewhere / not needed. You can make cake without flour, and people
> still call it a cake. You can make bread without yeast, and it is
> flatbread. You can make Linux without block device support, and it is
> still a kernel. Similarly, BRS-M is a reduced model, but it is still a
> recipe.

A text only mentioning "do not use flour" isn't a recipe.

Best regards

Heinrich

non-normative/recipes.adoc
andreiw: Ron, if you have a platform where you don't want ACPI and SMBIOS (i.e. you are using DT), then this use case fits squarely into the BRS-B (EBBR) profile. For BRS-B, ACPI and SMBIOS are optional. It's BRS-I where ACPI and SMBIOS are mandatory.

andreiw: > We believe that the firmware ecosystem is developed and should not be limited to UEFI alone. In this way, it can also adapt to more scenarios. From the BRS spec, there is no direct description of UEFI and ACPI binding. For example, coreboot can easily support ACPI/SMBIOS, but if you add EFI stub, it adds complexity and no better benefits.

FWIW, adding an alternate ACPI discovery mechanism itself adds complexity and no real benefit. It just shifts the problem elsewhere - namely to OSVs or cloud service providers. With an LBBR-like approach, you can still boot an unmodified Ubuntu or w/e installer/installation, with the stock kernel, etc. With your approach, you can't even boot anything unless you roll a custom kernel, or if your proposed changes are accepted and then if picked up by a distro (which is a lot of 'ifs').

A few weeks ago I asked the technical cost of implementing a minimal LBBR-like EFI stub in your firmware. I haven't seen any responses to that, but I do have personal experience doing so in an AArch64 U-boot fork about 10 years ago (e.g. EFI stub with dummy RT, used just to discover ACPI tables - this predates U-Boot's much more complete EFI implementation). The cost is insignificant - on the order of a half a day of work and 100 lines of code, if that. And the value is clear, especially when you avoid needless changes in the OSes you'd be booting that way. I can't imagine any kernel maintainer for a Linux distro agreeing that /another/ mechanism of kernel/firmware handoff is a good use of anyone's time. Feel free to loop those people in here from Canonical, Suse, Red Hat, Debian, NetBSD, etc, if I'm incorrect.

The purpose behind the BRS standard exist to make it easier to bring existing OSes and OS vendors to the nascent RISC-V ecosystem - it's a design choice to adopt known-good solutions instead of hand-rolling things unless there's really no alternative, to minimize tech debt incurred to support new systems and use-cases.

So far the only action item I see here is to continue to explore the CSP/cloud/LinuxBoot use-case. We should take LBBR as a start, again because it's already an accepted approach, and also because it doesn't suffer from NIH. Given our hopes around ratification schedule for BRS, I'd recommend getting a draft of something reviewable soon, else there's a good chance it won't make it into the first version of BRS.

andreiw:
> At the meeting yesterday we agreed that this could be tidied up by removing coreboot references, and it should be possible to get it in.

More than that. If your approach diverges from LBBR, then it needs a thorough review and sufficient buy-in - specifically from OS vendors. Yes - more attention than either BRS-I or BRS-B because these basically retrace the SBBR/EBBR steps. You've made a point - you want flexibility around how to construct firmware, and I think everyone acknowledged that and is willing to get the spec to cover this use-case (e.g for CSPs). Now please acknowledge our questions and concerns around prior art, technical debt, etc. "We feel our solution is better" isn't exactly grounds for something to be standardized, especially when the cost/benefit isn't a slam dunk.

jrtc27:
From a non-Linux perspective, even LBBR is pretty sucky. Yes, you get UEFI+DT/ACPI, so the actual OS boot part can work without modification, but it doesn't solve the bootloader story like EBBR does, since there's no standard UEFI boot services available. This means that FreeBSD has to build a variant of its bootloader as a Linux executable in order to support this, which is a bit perverse given it's not an OS-independent standard interface like UEFI is (imagine the uproar if WindowsBoot existed with some WBBR and in order to get a GRUB-like bootloader you had to build GRUB as a Windows executable).

andreiw:
> From a non-Linux perspective, even LBBR is pretty sucky. Yes, you get UEFI+DT/ACPI, so the actual OS boot part can work without modification, but it doesn't solve the bootloader story like EBBR does, since there's no standard UEFI boot services available. This means that FreeBSD has to build a variant of its bootloader as a Linux executable in order to support this, which is a bit perverse given it's not an OS-independent standard interface like UEFI is (imagine the uproar if WindowsBoot existed with some WBBR and in order to get a GRUB-like bootloader you had to build GRUB as a Windows executable). That or you forego the bootloader and lose out on the features it brings, like ZFS boot environments and module preloading.

LBBR exists because some amount of Cloud providers, who want to boot Linux bare metal, felt they wanted to codify their requirements in a spec to make it easier to share those requirements with Silicon providers. If you want to boot other  things, then your firmware ought to be BRS-B (EBBR) or BRS-I (SBBR) compliant.

jrtc27:
If it’s by Linux for Linux, why doesn’t Linux own that spec then?

ardbiesheuvel:
We have been having the same discussion in different venues now (Linux, RISC-V), but not in the right one (ACPI).

The arch maintainers in Linux (both arm64 and RISC-V) are reluctant to accept an alternative boot protocol that is not covered by an industry spec because they want to avoid ending up in the same situation as x86, which has so many boot protocols and variants that it is almost impossible to make substantial changes without breaking /something/.

Also, I will raise the same points here that have not been answered yet in any of the places where this has been discussed:
- how should the OS prioritize ACPI and DT if both are being used at the same time by the firmware to pass platform information to the OS;
- how should the OS infer memory attributes to be used my SystemMemory OpRegions created by AML (which lacks a facility to describe this, which is why Linux/arm64 and Linux/riscv cross reference the EFI memory map when creating memory mappings on behalf of ACPI AML code)

Personally, I think Linux could permit, e..g, /chosen/acpi-rsdp-addr to be used to pass the ACPI root table to the OS provided that
- OS is not booted via EFI
- DT must not carry anything except a /chosen node, and the OS must ignore any other nodes if it does;
- AML SystemMemory OpRegions must have MMIO device semantics, and accessing RAM is not permitted.

There are some remaining loose ends (e.g., Linux may make assumptions about PCI resource allocation when booting via ACPI) but we can discuss this further in the followup discussions.

The UEFI forum is the right venue to discuss these changes to the ACPI spec. Without such changes, I don't think Linux should (or will) consider patches that add support for non-EFI ACPI boot.

I am happy to discuss this further, but I cordially invite those in favor of adding these facilities to actually address the points that others and myself have raised. Anecdotal examples of x86 systems that are able to boot Linux in various different ways are not really relevant here.

andreiw:
If a reduced DT blob is used to convey the RSDP without UEFI, the OS also needs to know which memory ranges are used by firmware tables, to avoid trampling these. So you might need /reserved-memory. And ACPI doesn't really describe RAM (unless implied by something like SRAT entries), so you're probably needing at least one /memory node as well. So all in all you're parsing _DT_ and _ACPI_. What a mess.

ardbiesheuvel:
> If a reduced DT blob is used to convey the RSDP without UEFI, the OS also needs to know which memory ranges are used by firmware tables, to avoid trampling these. So you might need /reserved-memory. And ACPI doesn't really describe RAM (unless implied by something like SRAT entries), so you're probably needing at least one /memory node as well. So all in all you're parsing _DT_ and _ACPI_. What a mess.

Yeah, good point, I hadn't quite realized that. And this reinforces my point that this discussion needs to take place in the right venue, and not in a non-binding RISC-V spec PR thread.

This is another reason why the 'x86 can do it' argument does not hold here: Linux/x86 heavily relies on the E820 memory map internally, and it even converts the EFI memory map to E820 in the early boot code to address the impedance mismatch.

Non-x86 architectures such as RISC-V or arm64 have no pre-UEFI fallback like x86 did, and so everything that ACPI relies on needs to be reinvented when not booting via EFI. If this implies /memory and /reserved-memory nodes in DT along with ACPI tables for the hardware descriptions, I don't think Linux should go down this path. This means my suggestion to use the /chosen node and nothing else to permit ACPI without EFI is infeasible in practice, and should be disregarded.

andreiw:
> > If a reduced DT blob is used to convey the RSDP without UEFI, the OS also needs to know which memory ranges are used by firmware tables, to avoid trampling these. So you might need /reserved-memory. And ACPI doesn't really describe RAM (unless implied by something like SRAT entries), so you're probably needing at least one /memory node as well. So all in all you're parsing _DT_ and _ACPI_. What a mess.
> 
> Yeah, good point, I hadn't quite realized that. And this reinforces my point that this discussion needs to take place in the right venue, and not in a non-binding RISC-V spec PR thread.
> 

Entirely agreed. The BRS spec is also the wrong place to define such a novel mechanism. On a different note, so long as folks aren't bent on reinventing the wheel, I'm sure we could work on defining a CSP-friendlier firmware profile, based on actual CSP reqs/needs and existing well-defined technologies... the problem here is that we don't have the right crowd. There are practically no OSV, OEM or CSP voices in the BRS group. This doesn't prevent making _some_ progress (e.g. being inspired by similar work done in other ecosystems), but it means there's no real authority to come up with radically different approaches. You can't really trust Silicon companies to come up with things software or product people need.

weidongwd:
> > If a reduced DT blob is used to convey the RSDP without UEFI, the OS also needs to know which memory ranges are used by firmware tables, to avoid trampling these. So you might need /reserved-memory. And ACPI doesn't really describe RAM (unless implied by something like SRAT entries), so you're probably needing at least one /memory node as well. So all in all you're parsing _DT_ and _ACPI_. What a mess.
> 
> Yeah, good point, I hadn't quite realized that. And this reinforces my point that this discussion needs to take place in the right venue, and not in a non-binding RISC-V spec PR thread.
> 
> This is another reason why the 'x86 can do it' argument does not hold here: Linux/x86 heavily relies on the E820 memory map internally, and it even converts the EFI memory map to E820 in the early boot code to address the impedance mismatch.
> 
> Non-x86 architectures such as RISC-V or arm64 have no pre-UEFI fallback like x86 did, and so everything that ACPI relies on needs to be reinvented when not booting via EFI. If this implies /memory and /reserved-memory nodes in DT along with ACPI tables for the hardware descriptions, I don't think Linux should go down this path. This means my suggestion to use the /chosen node and nothing else to permit ACPI without EFI is infeasible in practice, and should be disregarded.

Why doesn't linux support memory nodes with ACPI in DT? I think passing memory map through memory nodes and passing ACPI tables through RSDP pointer is also a complete scheme.

weidongwd:
> At the meeting yesterday we agreed that this could be tidied up by removing coreboot references, and it should be possible to get it in.

Yes, I will remove coreboot references. The rest should merge into the BRS spec.

cuiyunhui-1:
> Also, I will raise the same points here that have not been answered yet in any of the places where this has been discussed:
> 
> * how should the OS prioritize ACPI and DT if both are being used at the same time by the firmware to pass platform information to the OS;

Currently in our FFI solution, the RSDP obtained through DT is filled into the acpi_arch_get_root_pointer() interface. The current logic is to obtain it through acpi_arch_get_root_pointer() first. If it cannot be obtained, go to the EFI entrance.

See drivers/acpi/osl.c

pa = acpi_arch_get_root_pointer();
if (pa)
return pa;

if (efi_enabled(EFI_CONFIG_TABLES)) {
if (efi.acpi20 != EFI_INVALID_TABLE_ADDR)
return efi.acpi20;
if (efi.acpi != EFI_INVALID_TABLE_ADDR)
return efi.acpi;
....

Does this mean there is no priority issue?

cuiyunhui-1:
> * how should the OS infer memory attributes to be used by SystemMemory OpRegions created by AML (which lacks a facility to describe this, which is why Linux/arm64 and Linux/riscv cross reference the EFI memory map when creating memory mappings on behalf of ACPI AML code)

FFI (firmware fdt interface) implements memmap in the following three points:
1. How to obtain the address of memmap?
Now the arm64 platform gets the struct efi_memory_map_data data in the fdt of efi_init() and then calls efi_memmap_init_early() to fill the efi.memmap.
FFI can also do this: get ffi.memmap from fdt in ffi_init(), because FFI itself is based on fdt.
Obviously, we no longer need arm64-like stubs.

2. How to fill the content of memmap?
Filling this memmap table requires bootloader to support the efi table component, and coreboot does not support efi related components, so memap can only be placed in dtb, and early_init_dt_scan_nodes() scans the memory node attribute of dtb in fdt.

After all, the arm64 memory segment is also obtained through early_init_dt_scan_nodes->early_init_dt_scan_memory().

3. How to use memmap?
It is all done through __acpi_get_mem_attribute()->efi_mem_attributes(). We can fill in ffi_mem_attributes() in the __acpi_get_mem_attribute() function. As for the priority: if efi_mem_attributes() fails, continue with ffi_mem_attributes().

As for the memory of the ACPI SMBIOS table itself, it needs to be reserved in early_init_dt_scan_memory(). This is something that all solutions must consider, not just FFI.

Have any questions about memmap?

cuiyunhui-1:
From the arm64 platform of efiboot, efi.memmap is obtained through fdt, the memory segment is also obtained from fdt, even efi_system_table is obtained through fdt, see efi_get_fdt_params(acpi/smbios are obtained through efi_system_table).

Oh, fdt plays a big role in arm64’s efiboot. In fact, this contradicts what you said before that acpi/efi and fdt are completely independent mechanisms. @ardbiesheuvel 


Why not just apply FFI? Why?

weidongwd:
Yunhui has answered some questions about Linux. Can his answer solve the above problems? If so, we should get this PR in.

ardbiesheuvel:
> From the arm64 platform of efiboot, efi.memmap is obtained through fdt, the memory segment is also obtained from fdt, even efi_system_table is obtained through fdt, see efi_get_fdt_params(acpi/smbios are obtained through efi_system_table).
> 
> Oh, fdt plays a big role in arm64’s efiboot. In fact, this contradicts what you said before that acpi/efi and fdt are completely independent mechanisms. @ardbiesheuvel
> 

DT is used internally in the Linux arm64 and RISC-V kernel, to pass information between the EFI boot stub and the kernel. This is not external API, and is subject to change. There are plans to share more common code between x86, RISC-V, arm64 and LoongArch, all of which use slightly different mechanisms now, which is a maintenance burden.

Using both DT and ACPI in parallel as external APIs to provide platform information to the OS is something completely different, and not something we should entertain IMHO.

cuiyunhui-1:
> > From the arm64 platform of efiboot, efi.memmap is obtained through fdt, the memory segment is also obtained from fdt, even efi_system_table is obtained through fdt, see efi_get_fdt_params(acpi/smbios are obtained through efi_system_table).
> > Oh, fdt plays a big role in arm64’s efiboot. In fact, this contradicts what you said before that acpi/efi and fdt are completely independent mechanisms. @ardbiesheuvel
> 
> DT is used internally in the Linux arm64 and RISC-V kernel, to pass information between the EFI boot stub and the kernel. This is not external API, and is subject to change. There are plans to share more common code between x86, RISC-V, arm64 and LoongArch, all of which use slightly different mechanisms now, which is a maintenance burden.
> 
> Using both DT and ACPI in parallel as external APIs to provide platform information to the OS is something completely different, and not something we should entertain IMHO.

Judging from the patches sent from upstream and the support for memmap, there are not many. Just fill in the entries for ACPI, SMBIOS, and memmap without adding more burden.

It will be done directly from the FDT rather than taking a roundabout way to wire to the EFI struct.

We hope not to follow a roundabout way on RISC-V, which makes it easier for developers to maintain, and we do not affect the original EFI logic.

weidongwd:
> > > From the arm64 platform of efiboot, efi.memmap is obtained through fdt, the memory segment is also obtained from fdt, even efi_system_table is obtained through fdt, see efi_get_fdt_params(acpi/smbios are obtained through efi_system_table).
> > > Oh, fdt plays a big role in arm64’s efiboot. In fact, this contradicts what you said before that acpi/efi and fdt are completely independent mechanisms. @ardbiesheuvel
> > 
> > 
> > DT is used internally in the Linux arm64 and RISC-V kernel, to pass information between the EFI boot stub and the kernel. This is not external API, and is subject to change. There are plans to share more common code between x86, RISC-V, arm64 and LoongArch, all of which use slightly different mechanisms now, which is a maintenance burden.
> > Using both DT and ACPI in parallel as external APIs to provide platform information to the OS is something completely different, and not something we should entertain IMHO.
> 
> Judging from the patches sent from upstream and the support for memmap, there are not many. Just fill in the entries for ACPI, SMBIOS, and memmap without adding more burden.
> 
> It will be done directly from the FDT rather than taking a roundabout way to wire to the EFI struct.
> 
> We hope not to follow a roundabout way on RISC-V, which makes it easier for developers to maintain, and we do not affect the original EFI logic.

What do you think of Yunhui's idea? If there is no doubt, can I submit a new PR that removes the coreboot description and then get it in?

andreiw:
> Historically, ARM has been incredibly accommodating of those projects. Dong Wei in particular has been very helpful with LinuxBoot and the LBBR. An original goal of LBBR, still of interest, is direct loading Linux from TF-A, bypassing UEFI entirely, so we can remove UEFI.
> 
> I'd like to see the RISC-V community be at least as accommodating as ARM. To put it another way, the software side of RISC-V should be as open as the hardware side to different uses and extensions.

We can for sure consider an LBBR approach, which as you well know, uses EFI data structures to pass ACPI/SMBIOS. Accommodating doesn't mean carte blanche adoption of every bespoke approach.  The LBBR approach involves no OS changes. You can boot an existing Linux distro that way. The ACPI via DT one clearly impacts OSVs. You don't like EFI, so others ought to somehow bend to your will. Have you asked these others? I mean the upstream kernel (Linux) and downstream distros, as well as non-Linux OSes.

CodingVoid:
> The ACPI via DT one clearly impacts OSVs.

> Using both DT and ACPI in parallel as external APIs to provide platform information to the OS is something completely different, and not something we should entertain IMHO.

Linux already uses DT and ACPI in parallel. Like mentioned above Linux internally uses the DT to transfer information from EFISTUB to the Kernel.

My suggestion would be that if the OS finds a reference to ACPI inside the DT it should just ignore DT and use ACPI instead (basically prioritize ACPI). I agree it doesn't make sense to pass ACPI and DT (which are essentially doing the same) to the OS. But in this case the DT is just a means to transfer the location of the ACPI tables. Still better than inventing a whole new format which would transfer the ACPI location to the OS. I mean Linux basically did the same. They didn't want to invent a new format to pass from EFISTUB to Kernel so they used FDT (I am guessing. I don't know the actual backstory.)

andreiw:
Closing as non-actionable.

CodingVoid:
> Closing as non-actionable.

Not sure what that is supposed to mean?

Issue #83, opened 2023-10-15
Define trap (interrupt/exception) handler delegation


orangecms:
At this point, it is more a less coincidental which trap handlers an SBI may or may not delegate to an OS.
While OpenSBI handles e.g. unaligned access, this is not to be expected in general.

Now comes the hard part:
Should we define a set of what _exactly_ is / should be delegated, or at least say that the OS should be _prepared_ for anything to be delegated? It is crucial for the OS to know. Otherwise, there are dozens of possible combos of (non-)delegations.

andreiw:
Can we start by enumerating the full list and the default behavior seen in OpenSBI?

adurbin-rivos:
> 
> I'd create a `trap-handlers.adoc` file and a link from the `sbi.adoc`. WDYT?

Sure. Go with that. However, I'm not sure it's really related to sbi per-se. It's more of how to configure m-mode delegation to s-mode, if m-mode is implemented.

ved-rivos:
In what case does it matter whether the interrupt/exception is delegated or is handled M-first? Should not the OS expect to handle all defined S exceptions and interrupts - whether delegated or emulated?

orangecms:
> In what case does it matter whether the interrupt/exception is delegated or is handled M-first? Should not the OS expect to handle all defined S exceptions and interrupts - whether delegated or emulated?

This is to make clear what to expect and what not, more elaborate:
https://github.com/riscv-non-isa/riscv-brs/pull/84#discussion_r1365900394

andreiw:
@ved-rivos @orangecms  can we revisit this and reach a conclusion?

orangecms:
> @ved-rivos @orangecms  can we revisit this and reach a conclusion?

That'd be great. I've literally just had another discussion around this the other day. While Linux now has the patches for handling unaligned access itself, they are currently not effective e.g. on JH7110 boards, where vendor firmware does not delegate the respective exceptions. So that potentially skews performance results, depending on the software.

andreiw:
Also @atishp04 

Do we want an SBI extension where an OS can negotiate delegations? I.e agree if a particular trap would go fw-only, fw-first or kernel-first.

That's probably more workable then attempting to pin down which traps can be delegated or not (or should be), as that's highly dependent on the system in question. It seems like it would also help solve the real issues you see with JH7110.

TobiasSchaffner:
[...]

PR #84, opened 2023-10-16
Add initial notes on trap handlers

orangecms:
initial draft on #83 

just give me a while, I'm not versed in ASCII doc yet

### DIFF
--- a/trap-handlers.adoc
+++ b/trap-handlers.adoc
@@ -1,0 +1,127 @@
+[[trap-handlers]]
+== Machine-Level Interrupts and Exceptions
+
+=== Delegation
+
+A RISC-V Machine mode execution environment may delegate handling interrupts and exceptions to the Supervisor or Hypervisor mode execution environment.
+
+For reference, see https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc#machine-cause-register-mcause[the definitions in the RISC-V ISA spec].
+
+==== Interrupts
+[%autowidth,float="center",align="center",cols=">,>,<",options="header",]
+|===
+| Code / Description |OpenSBI |oreboot SBI
+|0 _Reserved_ +
+1 Supervisor software interrupt +
+2 _Reserved_ +
+3 Machine software interrupt +
+4 _Reserved_ +
+5 Supervisor timer interrupt +
+6 _Reserved_ +
+7 Machine timer interrupt +
+8 _Reserved_ +
+9 Supervisor external interrupt +
+10 _Reserved_ +
+11 Machine external interrupt 
+12-15 _Reserved_ +
+&#8805;16 _Designated for platform use_ 
+| O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O
+| O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O +
+O
+|===
+
+X = explicitly set
+O = explicitly unset
+I = initial value unchanged
+
+==== Exceptions
+[%autowidth,float="center",align="center",cols=">,>,<",options="header",]
+|===
+| Code / Description |OpenSBI |oreboot SBI
+|0 Instruction address misaligned +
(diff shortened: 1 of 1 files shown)


trap-handlers.adoc
```diff
@@ -0,0 +1,127 @@
...
+=== Delegation
+
+A RISC-V Machine mode execution environment may delegate handling interrupts and exceptions to the Supervisor or Hypervisor mode execution environment.
+
+For reference, see https://github.com/riscv/riscv-isa-manual/blob/main/src/machine.adoc#machine-cause-register-mcause[the definitions in the RISC-V ISA spec].
+
+==== Interrupts
+[%autowidth,float="center",align="center",cols=">,>,<",options="header",]
+|===
+| Code / Description |OpenSBI |oreboot SBI
```
adurbin-rivos, line 13: Why reference m-mode implementations? Why not call out what the expected behavior should be? And we should probably include the reasoning.

orangecms: this is for reference now, based on this suggestion https://github.com/riscv-non-isa/riscv-brs/issues/83#issuecomment-1764583868

my pledge is to always delegate everything that S-mode could handle, which only excludes operations involving CSRs that are M-mode only, AIUI

andreiw: If I had implied we should go into describing M mode implementations, that was not the intention. The BRS spec describes S mode behavior as it relates to booting at OS. Some implementations may not even have an M mode (think VMs).

We should come up with expected behavior (or a mechanism of conveying the necessary info). But I kinda wanna second Ved's feedback or why we ought to care - I think the assumption is that the OS will take the exception eventually for reasons that it would care about (even if the original exception went to M, as it very well might, for example due to firwmare-first error handling)

andreiw: "my pledge is to always delegate everything that S-mode could handle" we're definitely into impl-specific choices, and I don't think the BRS spec is the place to push for those choices. Like with other architectures, work-arounds and additional complexity may end up getting put into M, but in a manner generally transparent to S mode software. What matters is not what's done, but what's the measurable impact.

orangecms: Gotcha. I had understood from this comment that BRS would supersede the platform spec:
https://github.com/riscv/riscv-platform-specs/pull/91#issuecomment-1707916908

So my intention was to get to a definition of behavior an S-mode OS would expect to avoid confusion where exceptions and/or interrupts are or are not arriving. That is what surprised me when Linux was not handling unaligned access traps, which is now being implemented in a pending patch series. Other OSes that are not part of that development history would benefit. Should this rather go into PRS, then? I would be happy to see it well-defined _somewhere_, not being picky where. BRS appeared to be closest to that from my current understanding. Where do you rather see it? It could also live in the wiki or in a blog post, I think.

For clarification: In which sense do you mean "transparent"?
There are two interpretations around:
1. opaque, the other component (S-mode here) is not aware of something
2. visible, the other component can notice and acknowledge

andreiw: Got it, so basically you want a base set of exceptions that an S-Mode OS can expect to be delivered? (like unaligned access).

Will bring this up in the BRS call today and come back with some guidance.

andreiw:
@orangecms @adurbin-rivos should revisit this.

adurbin-rivos:
> @orangecms @adurbin-rivos should revisit this.

Is there anything else to discuss?

andreiw:
@orangecms can you rework this PR to reference the still in progress FWFT extension as a required mechanism?

orangecms:
> @orangecms can you rework this PR to reference the still in progress FWFT extension as a required mechanism?
> 

Finally getting back to wrapping my head aroubd this.

So, in the end, I would remove the table/listing, and add the FWFT reference, correct?

andreiw:
I think this item is waiting on SBI 3.0, then we can reference the FWFT.

orangecms:
closing this since the proposed changes are opposing the original idea
see also https://github.com/riscv-non-isa/riscv-brs/issues/83#issuecomment-2725972597

PR #87, opened 2023-10-26
Add a new recipe to support non-UEFI firmware.

weidongwd:
The second version was submitted, removing the coreboot description based on PR72.


andreiw:
So regardless of the benefit of the proposal, you insist on not providing the necessary details to make it a spec. I don't see any actual DT bindings nor any discussion of what the DT looks like (e.g. wrt memory reservations, for example).

Meanwhile LBBRv2 was announced at OCP. Have you taken a look at that?

andreiw:
Closing as non-actionable.

PR #90, opened 2023-11-28
Smbios cleanup

andreiw:
As discussed in the PRS call, here's my rework on the SMBIOS bits that were left orphaned over at riscv-smbios.

### DIFF
--- a/smbios.adoc
+++ b/smbios.adoc
@@ -40,15 +40,13 @@ architecture 07h (64-bit RISC-V) only.
 | Offset | Version | Name      | Length | Value   | Description
 | 00h| 0100h|Revision|WORD|Varies|See <<smbios-psd-ver>>.
 | 02h| 0100h| Hart ID| QWORD| Varies| The ID of this RISC-V hart
-| 0Ah| 0100h| Boot hart| BYTE| Boolean| 1: This is the boot hart. +
-0: This is not the boot hart.
-| 0Bh| 0100h| Machine Vendor ID | QWORD| Varies| The vendor ID of this
+| 0Ah| 0100h| Machine Vendor ID | QWORD| Varies| The vendor ID of this
 RISC-V hart
-| 13h| 0100h| Machine Architecture ID| QWORD| Varies| Base
+| 12h| 0100h| Machine Architecture ID| QWORD| Varies| Base
 microarchitecture of the hart. Value of 0 is possible to indicate the field is
 not implemented. The combination of Machine Architecture ID and Machine Vendor
 ID should uniquely identify the type of hart microarchitecture that is implemented.
-| 1Bh| 0100h| Machine Implementation ID| QWORD| Varies| Unique encoding
+| 1Ah| 0100h| Machine Implementation ID| QWORD| Varies| Unique encoding
 of the version of the processor implementation. Value of 0 is possible to indicate
 the field is not implemented. The Implementation value should reflect the design of
 the RISC-V processor and not the surrounding system.
--- a/smbios.adoc
+++ b/smbios.adoc
@@ -72,17 +72,6 @@ handler in Machine-Mode for the certain exceptions through environment call.
 Bit set to one means the corresponding interrupt is delegated to supervisor execution
 environment. Otherwise, supervisor execution environment must register the event
 handler in Machine-Mode for the certain interrupts through environment.
-| 38h| 0100h| The register width (XLEN)| BYTE| ENUM| The width of register
-supported by this RISC-V Hart
-| 39h| 0100h| Machine Mode native base integer ISA width (M-XLEN)| BYTE| ENUM|
-The width (See below) of Machine Mode native base integer ISA supported by this RISC-V
-Hart
-| 3Ah| 0100h| Reserved| BYTE| ENUM| Placeholder for Hypervisor Mode
-| 3Bh| 0100h| Supervisor Mode native base integer ISA width (S-XLEN)| BYTE| ENUM|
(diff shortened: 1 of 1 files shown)


smbios.adoc
```diff
@@ -46,50 +46,49 @@ The current version of the structure is 000Ah (v0.10).
 Bit 7:0 Minor revision +
 The newer revision of RISC-V Processor-specific Block
 Structure is backward compatible with older version of this structure.
-|02h|000Ah|Structure Length|Byte|110|Length of Processor-specific Data
-| 03h| 000Ah| Hart ID| DQWORD| Varies| The ID of this RISC-V Hart
-| 13h| 000Ah| Boot Hart| BYTE| Boolean| 1: This is boot hart to boot system +
+| 02h| 000Ah| Hart ID| DQWORD| Varies| The ID of this RISC-V Hart
```
adurbin-rivos: This comment is more about the version. Should we bump the version since we're changing fields?

smbios.adoc
```diff
@@ -29,67 +29,79 @@ on top of cite:[SMBIOS], and is optional and recommended for BRS-B. Additional n
...
 The width (See below) of Machine Mode native base integer ISA supported by this RISC-V
 Hart
-| 3Ah| 000Ah| Reserved| BYTE| ENUM| Placeholder for Hypervisor Mode
-| 3Bh| 000Ah| Supervisor Mode native base integer ISA width (S-XLEN)| BYTE| ENUM|
+| 3Ah| 0100h| Reserved| BYTE| ENUM| Placeholder for Hypervisor Mode
+| 3Bh| 0100h| Supervisor Mode native base integer ISA width (S-XLEN)| BYTE| ENUM|
 The width (See below) of Supervisor Mode native base integer ISA supported by this RISC-V
 Hart
-| 3Ch| 0000Ah| User Mode native base integer ISA width (U-XLEN)| BYTE| ENUM| The
+| 3Ch| 00100h| User Mode native base integer ISA width (U-XLEN)| BYTE| ENUM| The
```
adurbin-rivos: an extra leading 0

smbios.adoc
```diff
@@ -29,67 +29,79 @@ on top of cite:[SMBIOS], and is optional and recommended for BRS-B. Additional n
...
+The minor revision is to be incremented anytime new fields are added in a backwards-compatible manner. The major revision is to be incremented on backwards-incompatible changes.
+
+[cols="1,1,1,1,3", width=95%, align="center", options="header"]
+|===
+| Version | Bits 15:8+
+Major revision
+| Bits 7:0+
+Minor revision
+| Combined | Description
+| v1.0 | 01h | 00h | 0100h | First BRS-defined definition
```
adurbin-rivos, line 71: My other comment is now addressed.  Should we put this first at the beginning of the section like it was before?

smbios.adoc
```diff
@@ -22,7 +22,51 @@ on top of cite:[SMBIOS], and is optional and recommended for BRS-B. Additional n
...
+| 02h| 0100h| Hart ID| QWORD| Varies| The ID of this RISC-V hart
+| 0Ah| 0100h| Boot hart| BYTE| Boolean| 1: This is the boot hart. +
+0: This is not the boot hart.
+| 0Bh| 0100h| Machine Vendor ID | QWORD| Varies| The vendor ID of this
+RISC-V hart
+| 13h| 0100h| Machine Architecture ID| QWORD| Varies| Base
+microarchitecture of the hart. Value of 0 is possible to indicate the field is
+not implemented. The combination of Machine Architecture ID and Machine Vendor
+ID should uniquely identify the type of hart microarchitecture that is implemented.
+| 1Bh| 0100h| Machine Implementation ID| QWORD| Varies| Unique encoding
```
vlsunil, line 51: While I understand this is a spec, I have an implementation question for these fields. These vendor ID/impl ID for each hart can be fetched using SBI call from S-mode FW only from the hart which executes it right? Typically, the SMBIOS creator FW like EDK2 runs with single CPU. How is it supposed to create these fields  each hart?

smbios.adoc
```diff
@@ -22,7 +22,51 @@ on top of cite:[SMBIOS], and is optional and recommended for BRS-B. Additional n
...
+The structure is defined in a manner consistent with the DMTF specification
+language (cite:[SMBIOS]), and is valid for processors declared as
+architecture 07h (64-bit RISC-V) only.
+
+[cols="2,2,3,2,2,4", width=95%, align="center", options="header"]
+|===
+| Offset | Version | Name      | Length | Value   | Description
+| 00h| 0100h|Revision|WORD|Varies|See <<smbios-psd-ver>>.
+| 02h| 0100h| Hart ID| QWORD| Varies| The ID of this RISC-V hart
+| 0Ah| 0100h| Boot hart| BYTE| Boolean| 1: This is the boot hart. +
```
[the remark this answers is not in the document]
andreiw: I would agree with you that it isn't. I'll take it out.

smbios.adoc
andreiw: Personally I think its fine where it is. There's a link to the section from the table so you can easily jump to it.

smbios.adoc
andreiw: These fields are ultimately deleted in the patch set.

adurbin-rivos (approved):

vlsunil (approved):
Looks good to me. Thanks a lot for taking care of this.

Now, this spec being moved to BRS, an ECR to SMBIOS spec required to update the link it is pointing for table 44?

PR #93, opened 2023-12-04
Add a few generic uefi secure boot requirements that don't make sense…

andreiw:
… to fold into the server platform spec

These came from looking over the BBSR.


andreiw:
Fixed the one nit found (a least -> at least)

adurbin-rivos (approved):

PR #126, opened 2024-03-01
Brs b no pci required

rminnich:
End-users have had trouble over the years with requiring BIOS to
configure PCIe. On the System 76 laptop with 4 thunderbolt
ports, it was an issue; and, on hyperscalar server platforms,
there have been cases where it is best to let linux configure
pcie, as firmware does not optimize to suit the case. The
system-76 case was a case also seen years ago on HPC nodes: BIOS
was required to configure the system for maximum generality,
making certain hardware impossible to use (4th thunderbolt port
on system 76; Quadrics NICs on HPC); in each case, the fix was
to get the BIOS to do minimal PCI configuration. Coreboot has a
build time option to enable this minimal PCI configuration,
deveoped jointly by Google and Intel.

### DIFF
--- a/uefi.adoc
+++ b/uefi.adoc
@@ -45,7 +45,7 @@ See additional <<uefi-rt, requirements under UEFI Runtime Services>>.
 [%header, cols="5,25"]
 |===
 | ID#     ^| Requirement
-| UIO_010 | Systems implementing PCIe MUST always initialize all root complex hardware and perform resource assignment for all endpoints and usable hotplug-capable switches in the system, even in a boot scenario from a non-PCIe boot device.
+| UIO_010 | BRS-I Systems implementing PCIe MUST always initialize all root complex hardware and perform resource assignment for all endpoints and usable hotplug-capable switches in the system, even in a boot scenario from a non-PCIe boot device. For BRS-B systems, this implementation is optional.
 2+| _This is a stronger requirement than the PCI Firmware Specification firmware/OS device hand-off state (cite:[PCIFW] Section 3.5). <<uefi-guidance-pcie, See additional guidance>>._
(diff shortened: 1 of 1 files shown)


rminnich:
I'm definitely doing this against the wrong branch; help requested.

adurbin-rivos:
> I'm definitely doing this against the wrong branch; help requested.

[riscv-non-isa:spec_changes](https://github.com/riscv-non-isa/riscv-brs/tree/spec_changes) from [rminnich:brs-b-no-pci-required](https://github.com/rminnich/riscv-brs/tree/brs-b-no-pci-required)

I think you are targeting the wrong branch, 'tree/spec_changes', in the merge request which is seeding a new branch.

rminnich:
OK, try 2, I'm supposed to pull against main, right? Don't know how I got that confused, sorry.

adurbin-rivos:
@andreiw  I'm ok with this change as it doesn't fundamentally change anything given that BRS-B can do what it desires while maintaining the BRS-I equivalent.

andreiw:
Hey folks.

Agreed as per mailing list discussion. However, _everything_ under the UEFI section is optional for BRS-B. So I think we _may_ need to make that more obvious, but adding a BRS-B disclaimer just to UIO_010 is going to be more confusing (as it might imply other requirements are somehow mandatory for BRS-B).

A

rminnich:
I think both these suggestions are better than this PR, once you have a PR up I'll close this one. Thanks.

andreiw:
Ron - https://github.com/riscv-non-isa/riscv-brs/issues/128 is up

andreiw:
Resolved #128 via https://github.com/riscv-non-isa/riscv-brs/pull/130

So I think we should be good here.

Issue #127, opened 2024-03-05
ACPI Namespace devices for APLIC and PLIC


vlsunil:
NOTE: This is still undecided and discussion in progress, just raising the issue here to keep track of it. We can close this if not required.

As part of upstreaming AIA drivers and ACPI support for AIA and PLIC, linux driver model has moved to probe all non-root drivers as platform drivers with FW defining the probe order. Especially APLIC can not be probed early since it needs whole platform device infrastructure to setup MSI device domains. Maintainers are not willing to support core_initcall() for these drivers since they are fragile. So,  creating the dependency between APLIC/PLIC interrupt controllers  and the actual device is the only way to ensure probing order. This would mean something like adding _DEP for devices on APLIC/PLIC. 

So, to add the dependency proposal is to mandate below things in BRS.

1) FW should add namespace devices for APLIC(RSCV0001) and PLIC (RSCV0002) along with MADT. These devices should support _GSB to map correct MADT structure.
2) Regular platform devices should add Interrupt ResourceSource to corresponding APLIC/PLIC for interrupts.
3) PNP devices like 16550A (PNP0501) should add _DEP on corresponding APLIC/PLIC.

vlsunil:
Correct. I still think MADT is alone sufficient. Created this issue as per Andrei's suggestion to keep track.

> I guess this is a follow up to your message on this thread. https://lore.kernel.org/lkml/20231226235602.GA1483795@bhelgaas/T/#mcc7c15fcc38042cfbb81f38f7b3220312360acbc
> 
> As per my understanding from the thread, you still prefer the current approach on v3. But if we have to adopt Rafael's suggestion, we have to do as suggested above. Please let me know if I misunderstood something

Correct. I still think MADT is alone sufficient. Created this issue as per Andrei's suggestion to keep track.

andreiw:
So for the moment being, no changes are needed to the BRS spec. Let's keep this issue open for now.

andreiw:
This is a current work item for the spec - not future.

andreiw:
And we're merged.

PR #143, opened 2024-04-11
ACPI: Add requirement to have PLIC/APLIC namespace devices

vlsunil:
PLIC and APLIC need to be namespace devices as well so that devices can add dependencies using _DEP. This is required for the OS to ensure drivers are probed in proper order.

So, this commit
1) Creates a new section under ACPI to list of ACPI IDs maintained for RVI standard devices.
2) Adds new requirement to mandate namespace devices for PLIC and APLIC. 3) Mandates _GSB required to map the namespace device to MADT. 4) Mandates _DEP to indicate dependency between devices and interrupt controller for GSI interrupts.

This is discussed in PRS meeting on 04/11/2024. This should close https://github.com/riscv-non-isa/riscv-brs/issues/127

PoC available at:
https://github.com/vlsunil/qemu/tree/riscv_acpi_dep_v1
https://github.com/vlsunil/linux/tree/acpi_b2_v4_dep_v1

### DIFF
--- a/acpi-id.adoc
+++ b/acpi-id.adoc
@@ -1,0 +1,30 @@
+[[acpi-ids]]
+=== RVI specific ACPI IDs
+
+ACPI ID is used in the _HID (Hardware ID), _CID (Compatibility ID) or
+_SUB (Subsystem ID) objects as described in the ACPI Specification for
+devices that do not have a standard enumeration mechanism. The ACPI ID
+consists of two parts: a Vendor ID followed by a product identifier.
+
+Vendor IDs consist of 4 characters, each character being either an
+uppercase letter (A-Z) or a numeral (0-9). The Vendor ID should be
+unique across the Industry and registered by the UEFI forum. For RVI
+standard devices, **"RSCV"** is the Vendor ID registered. Vendor-specific
+devices can use an appropriate Vendor ID registered for the manufacturer.
+
+Product Identifiers are always four-character hexadecimal numbers (0-9
+and A-F). The Device Manufacturer is responsible for assigning this
+identifier to each product model.
+
+This document contains the canonical list of ACPI IDs for the namespace
+devices that adhere to the RVI specifications. The RVI task groups may
+make pull requests against this repository to request the allocation of
+ACPI ID for any new device.
+
+[width=100%]
+[%header, cols="5,25"]
+|===
+| ACPI ID     ^| Device
+| RSCV0001     | RISC-V Platform-Level Interrupt Controller (PLIC)
+| RSCV0002     | RISC-V Advanced Platform-Level Interrupt Controller (APLIC)
+|===
--- a/acpi.adoc
+++ b/acpi.adoc
@@ -35,6 +35,11 @@ available to an OS loader via the standard UEFI EFI_GRAPHICS_OUTPUT_PROTOCOL int
  ** Use Interface Type 0x12 (16550-compatible with parameters defined in Generic Address Structure).
  ** There must be a matching AML device object.
 2+| _<<acpi-guidance-spcr, See additional guidance>>_.
+| [[acpi-namespace-dev]]ACPI_080 | Depending on the interrupt controller
+implemented by the system, PLIC or APLIC namespace devices must be
+present in the ACPI namespace along with MADT entries. <<acpi-ids,
+See RVI ACPI IDs>>.
+2+| _Also see <<acpi-irq-gsb, AML_090>> and <<acpi-irq-dep, AML_100>>_.
 |===
 
 [[acpi-aml]]
@@ -69,4 +74,14 @@ vendor-specific OS driver for correct function (SPI, I2C, etc), the TAD must
 be functional if the OS driver is not loaded. That is, when a dependent
 driver is loaded, an AML method switches further accesses to go
 through the driver-backed OperationRegion._
+| [[acpi-irq-gsb]] AML_090 | PLIC and APLIC devices must support
+Global System Interrupt Base (_GSB, cite:[ACPI] Section 6.2.7).
+<<acpi-guidance-gsi-namespace, See additional guidance>>.
+| [[acpi-irq-dep]] AML_100 | Devices which use Global System
+Interrupts (GSI) must indicate the dependency on corresponding
+interrupt controller using Operation Region Dependencies (_DEP,
+cite:[ACPI] Section 6.5.8).
+<<acpi-guidance-gsi-namespace, See additional guidance>>.
 |===
+
+include::acpi-id.adoc[]
--- a/non-normative/acpi.adoc
+++ b/non-normative/acpi.adoc
@@ -104,6 +104,41 @@ not be decoded in a system-specific manner to divine CPU topology.
 The PPTT Processor Properties Topology Table (PPTT) is to be used to
 describe affinities.
 
+[[acpi-guidance-gsi-namespace]]
+==== PLIC/APLIC as namespace devices
+The devices which have wired interrupts (GSI), should provide the
+dependency on the appropriate interrupt controller similar to below
+example with PLIC.
+
+
+ Scope (\_SB)
+ {
+    Device (IC00)
+    {
+        Name (_HID, "RSCV0001") // _HID: Hardware ID
+        Name (_UID, Zero)  // _UID: Unique ID
+        Method(_GSB) {
+            Return (0x10) // Global System Interrupt Base for this PLIC starts at 16
+       }
+    }
+    Device (DEV1)
+    { 
+      ...
+      Name (_DEP, Package() {\_SB.IC00})
+      Name (_CRS, ResourceTemplate ()  // _CRS: Current Resource Settings
+      {   
+          Memory32Fixed (ReadWrite,
+            0x10010000,         // Address Base.
+            0x00010000,         // Address Length
+            )
+          Interrupt (ResourceConsumer, Level, ActiveHigh, Exclusive, ,,)
+          { 
+            0x10,
+          }
+      })
+    }
+ }
+
 [[acpi-guidance-pcie]]
 ==== PCIe
 


vlsunil:
@andreiw @atishp04 @xypron : Please help reviewing this.

acpi.adoc
```diff
@@ -64,9 +69,16 @@ objects.
 | [[acpi-tad]] AML_070 | Systems with a Real-Time Clock on an OS-managed bus (e.g. I2C, subject to arbitration issues due to access to the bus by the OS) MUST implement the Time and Alarm Device (TAD).
 2+| _Also see <<uefi-rtc, URT_020>>_.
 | AML_080 | Systems implementing a TAD must be functional without additional system-specific OS drivers.
+| [[acpi-irq-gsb]] AML_090 | PLIC and APLIC devices must support
+**_GSB** method. <<acpi-ids, See RVI ACPI IDs>>.
```
andreiw, line 73: Suggest this as being _GSB (Global System Interrupt Base), with no need for bolding as we haven't done that elsewhere

andreiw: Even better would be something like Global System Interrupt Base (_GSB, cite:[ACPI] Section 6.2.7).

acpi.adoc
```diff
@@ -64,9 +69,16 @@ objects.
...
+| [[acpi-irq-dep]] AML_100 | Devices which use Global System
+Interrupts (GSI) must indicate the dependency on corresponding
+interrupt controller using **_DEP** method.
```
andreiw, line 76: Not a comment on your text but... this is still just so weird. Have you taken this to ASWG? This is ACPI, Device Tree. We can't just make up stuff just because someone in Linux thinks it makes sense. There's no reason ACPI on RISC-V has to be done in (what looks like) an entirely crazy fashion. This is just so....alien to ACPI on every architecture out there. Why can't the OS take an implicit dependency, then look up the IC controller based on GSI range in the MADT entry?

andreiw (changes requested):
First of all thanks for the PR. I've left comments on the the actual adoc contribution, which are trivial in nature (please build and make sure it looks ok).

While not a blocker for the contribution proper, I'd like to understand why on earth ACPI on RISC-V has to be different from ACPI on anything else, and whether we have ASWG consensus any of this being a good idea. Linux kernel maintainers are not ASWG and it's not clear the right level of ACPI expertise is being applied.

acpi.adoc
andreiw: @jmcneill-vmware your feedback on this?

acpi-id.adoc
```diff
@@ -0,0 +1,30 @@
...
+This document contains the canonical list of ACPI IDs for the namespace
+devices that adhere to the RVI specifications. The RVI task groups may
+make pull requests against this repository to request the allocation of
+ACPI ID for any new device.
+
+[width=100%]
+[%header, cols="5,25"]
+|===
+| ACPI ID     ^| Device
+| RSCV0001     | RISC-V Platform-Level Interrupt Controller (PLIC)
```
andreiw, line 28: Can you add a section to the non-normative ACPI guidance on examples of Device (...) definitions for _both_ of these?

acpi.adoc
andreiw: Also, if there's an AML-centric mechanism adopted for interrupt controller discovery, do we still need the relevant MADT entries? What value would those provide if they are not used by Linux?

acpi-id.adoc
vlsunil: Sure

acpi.adoc
```diff
@@ -34,6 +34,11 @@ available to an OS loader via the standard UEFI EFI_GRAPHICS_OUTPUT_PROTOCOL int
  * For NS16550-compatible UARTs:
  ** Use Interface Type 0x12 (16550-compatible with parameters defined in Generic Address Structure).
  ** There must be a matching AML device object.
+| [[acpi-namespace-dev]]ACPI_080 | Depending on the interrupt controller
+implemented by the system, PLIC or APLIC namespace devices must be
+present in the ACPI namespace along with MADT entries. <<acpi-ids,
+See RVI ACPI IDs>>.
+2+| _Also see <<acpi-irq-gsb, AML_090>> and <<acpi-irq-dep, AML_100>>_.
```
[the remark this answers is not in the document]
vlsunil: Oops, sorry. Let me correct it.

acpi.adoc
vlsunil: Sure

acpi.adoc
vlsunil: Thanks Andrei. Good questions.

The issue is primarily with APLIC. Wired-to-MSI bridges are kind of new class of devices. Other architectures support either MSI or wired. The need for this type of device came in ARM fairly recently (compared to the existence of the CPU architectures) with MBIGEN device from HiSilicon and framework support was added to support such class of interrupt controllers. (REF: https://lwn.net/Articles/660893/).

The question of probing order came up with MBIGEN as well. (REF: https://lore.kernel.org/linux-arm-kernel/563B22F1.901@arm.com/). Obviously, using EPROBE_DEFER in all possible drivers will get profound NAK. So, people came up with fw_devlink framework to resolve these issues and used in DT. You can see that ACPI support for MBIGEN was also added later (https://yhbt.net/lore/all/58D974D2.7060809@linaro.org/) which is also a namespace device (HISI0152). The probing order issue in ACPI was resolved using ResourceSource since it doesn't support fw_devlink.

As I can see, MBIGEN is used only with limited set of devices for a platform. So, using ResourceSource will work. Other devices can still continue to use the wired interrupt space in GIC.  But APLIC is more generic which needs to be supported for any device using Wired interrupts including PCI INTx pins. There are many devices/drivers (ex: PNP, GED etc) which are probed differently in ACPI assuming interrupt controllers are probed early. If such devices are required to be supported with MBIGEN, I am sure they will have the same issue as APLIC. So, unless probing of such devices itself is delayed, it is going to cause issue one way or other. Hence, _DEP is recommended which is simple to use.

Using _DEP for this type of ensuring probing order is not new. For ex: https://patchwork.kernel.org/project/linux-media/patch/20211203102857.44539-2-hdegoede@redhat.com/.

Regarding the question of MADT and namespace device: Namespace devices need more details. It can either support _MAT or lookup MADT directly. MBIGEN since it is not a standard architecture interrupt controller, it used _DSD which IMO we should avoid. The reason I kept MADT is, some other OS may be able to use MADT and ignore namespace devices.
So, should we make that either  _MAT or MADT ?

andreiw: Thanks for the in-depth explanation!

Can we avoid the _DEP for PLIC systems, then, since these are wire interrupts and not using wire-to-msi? That way it's obvious the _DEP is for a reason, not a blanket policy just for RV.

Using _DEP is not necessarily new, but taking a mandatory _DEP in every Device (...) with an interrupt resource most surely is. And you have to know which one you have to _DEP on, which is going to be a trainwreck IMHO. Can't the OS synthesize an implicit "GSI interrupt" device, which every ACPI device can depend on, that will handle the probing and initialization of the APLIC (based on MADT, based on Device()s with a certain HID, w/e)?

( The MBIGEN approach doesn't raise any questions in my mind. Those aren't GSIs. It's using the correct mechanism, ResourceSource, to deal with these not-GSI interrupts. )

Like I said, not a gate for this PR at this point, but at the end of the day this is Linux-specific implementation leaking into a spec. Do we have buy-in from the ACPI spec body - the ASWG?

My other worry is if interrupt controller initialization is moved entirely to depend on AML nodes, then the "de facto" OS (Linux) is not going to use the MADT entries. This will mean this to become an untested path, and likely filled with garbage that other OSes will just trip on.

acpi.adoc
andreiw: So if we move away from MADT entries, then these devices would also return _MAT, right?

acpi.adoc
atishp04, line 76: We discussed this in PRS TG. This should be optional. Other OS may not have the same issue. Firmware doesn't need to specify _GSB and _DEP in that case. Isn't it ?

atishp04: As per our discussion in the PRS TG, this has not discussed in the ASWG yet. I think it would be good to just discuss the problem space and propose the solutions to see if there is a better way to do things. I understand Linux ACPI/irq maintainers doesn't attend ASWG but if there is a better way proposed in ASWG, we may be able to convince the ACPI maintainers.

acpi.adoc
vlsunil: _MAT is not really mandatory. Even IOAPIC namespace device don't need _MAT. OS will look at the MADT entry directly using _GSB.

acpi.adoc
vlsunil: > Thanks for the in-depth explanation!
> 
> Can we avoid the _DEP for PLIC systems, then, since these are wire interrupts and not using wire-to-msi? That way it's obvious the _DEP is for a reason, not a blanket policy just for RV.
> 
I thought it will be more confusing for people when to use _DEP if we make it required only for APLIC because APLIC also has direct mode.

> Using _DEP is not necessarily new, but taking a mandatory _DEP in every Device (...) with an interrupt resource most surely is. And you have to know which one you have to _DEP on, which is going to be a trainwreck IMHO. 
>
I share the same concern, but this situation needs a new solution IMO. It can be either in OS or FW.  I think these discussions would eventually lead to some good solution.

>Can't the OS synthesize an implicit "GSI interrupt" device, which every other ACPI device can depend on, that will handle the lazy probing and initialization of the APLIC (based on MADT, based on Device()s with a certain HID and the _GSB, w/e)?
>
That is fw_devlink approach I tried which is not getting traction from ACPI maintainer. Also, _PRT is another way dependency exists which makes it tricky to use this approach. _DEP has advantage that it ensures scan itself is deferred if dependencies are not met.
 
> Can the _DEP target a "container" device that contains all the other devices? Can the _DEP be targeted towards a /_single_/ Device() object, which can contain the actual APLIC children with their _GSBs? How do we make this not seem like a bunch of crutches or be terrible to implement/maintain for IFVs?
> 
Hmm, I will have to think about it. IIUC, you are trying to avoid mapping exact interrupt controller for a device. But that kind of mapping he has to do any way with GSIs no?
> ( The MBIGEN approach doesn't raise any questions in my mind. Those aren't GSIs. It's using the correct mechanism, ResourceSource, to deal with these not-GSI interrupts. The ResourceSource handling in a kernel may mean lazy binding. )
> 
> Like I said, not a gate for this PR at this point (it's exactly the kind of short term solution that makes sense to put into the BRS), but at the end of the day this is a Linux-specific implementation issue leaking into a spec. Do we have buy-in from the ACPI spec body - the ASWG?
> 
I submitted this PR because of BRS deadline of ARC review. How much time do we have or can we keep this open item and close as part of ratification?
Let me start a mail thread with ASWG and copy Rafael.

> My other worry is if interrupt controller initialization is moved entirely to depend on AML nodes, then the "de facto" OS (Linux) is not going to use the MADT entries. This will mean this to become an untested path, and likely filled with garbage that other OSes will just trip on. Keeping the same info in both places is not a good idea.
Not really. Like I said, OS will look at MADT entries itself and _MAT is not  required. IOAPIC works in same way in linux. Namespace device is required only to ensure probe ordering.

vlsunil: Yeah, but FW will be common for different OSs. So I think even if one OS needs it, it will become mandatory for firmware. I was initially thinking keeping it bit open saying "Devices should indicate the dependency appropriately" without mentioning any ways like _DEP. But I guess that might cause confusion as well.

andreiw: We can take this PR into the BRS. This follows our policy of taking in the things we need to make progress instead of waiting on the upstream to fix itself. All I'm saying is that it's super important to seek alignment with ASWG instead of ending up defining our own ACPI flavor.

It should be safe to add the _DEP requirement here. Worst case (or best case, I suppose, depending on how you look at it),  the requirement gets deprecated in the future but has no ill effect on implementations.

andreiw: Yeah that wouldn't be really actionable, because "indicating appropriately" sounds like being able to enumerate every crutch needed by every OS kernel out there :-).

acpi-id.adoc
vlsunil: I added an example for PLIC. I didn't see need to repeat the same thing for just HID change. Let me know if I should.

acpi.adoc
vlsunil: Sure. Let me try adding dependencies automatically without needing explicit _DEP. If the maintainer agrees, we can remove this _DEP requirement, just keep the namespace device + _GSB. I hope to conclude before freeze deadline.

acpi-id.adoc
andreiw: This is good.

andreiw (approved):

Issue #147, opened 2024-04-15
smbios: Define processor ID as SoC ID


andreiw:
As you know, RISC-V defines the mvendorid, marchid and mimpid CSRs which are supposed to help with identifying the implementation.

Short synopsis of what these are about:
- The mvendorid CSR is a 32-bit read-only register providing the JEDEC manufacturer ID of the provider of the core (or 0 if you are not commercial)
- The marchid CSR is an MXLEN-bit read-only register encoding the base microarchitecture of the hart. Non-commercial IDs are allocated globally by RVI, while commercial IDs are specific to the mvendorid.
- The mimpid CSR provides a unique encoding of the version of the processor implementation. This register must be readable in any implementation, but a value of 0 can be returned to indicate that the field is not implemented. The Implementation value should reflect the design of the RISC-V processor itself and not any surrounding system.

So mvendorid and marchid are tied together. You need to know the mvendorid to correctly parse the marchid. So mvendorid and marchid *uniquely* identify the hart microarchitecture and vendor. It's unhelpful that there's no breakdown for "marchid". Like... where does the errata revision go? If that's vendor specific, then that's unfortunate (and where is the RVI definition for the non-commercial cores?).

Mimpid... gives me a headache. It's unclear what the heck a "processor implementation is". Is the "processor" the hart, the physical package or can it include surrounding IP in a SoC-like design? But if mimpid is the processor/SoC what is the processor/SoC vendor? Mvendorid can't be that - that's the hart IP vendor. Consider, say, an NXP LS chip using Cortex-A72 cores. The cores are made by Arm. The SoC is made by NXP. There can be erratas in how the CPU IP is integrated, and errata in related surrounding IP, so it's important to know the SoC vendor. Or consider a Ventana chiplet being part of someone's SoC design. Clearly there can be Ventana errata (that you would want to be able to identify using mvendor/marchid) and there can be integration errata relevant only to the SoC vendor/implementation.

Ok so we still don't know what mimpid is, but it seems like we do want a SoC ID.

I propose adding a SoC ID field to our Type 44 Processor-Specific Data for SMBIOS. This would be a JEDEC code for the vendor, a vendor specific identifier and an additional vendor-specific version field.

Thoughts?

andreiw:
Daniel M said:

[...]

PR #148, opened 2024-04-15
smbios: update processor ID for RISCV to reflect the SoC ID

andreiw:
Implements https://github.com/riscv-non-isa/riscv-brs/issues/147

The way this is done, it's forwards compatible with the DMTF spec text, which should help with upstreaming. Compatible in the sense that the "old" way just provided the JEP-106 value in the first DWORD (and had 0 for the second DWORD).

Note: unlike Arm, don't bother with a processor characteristic, as there is no analogue to the SMCCC SoC ID call.

### DIFF
--- a/smbios.adoc
+++ b/smbios.adoc
@@ -40,12 +40,17 @@ language (cite:[SMBIOS]).
 [[smbios-type04]]
 === Type 04 Processor Information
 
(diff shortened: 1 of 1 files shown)


adurbin-rivos:
Seems straight forward enough. It's basically informative. If we are expecting to do driver probing matches on this, we probably want more rigor.

Issue #24, opened 2023-03-30
Dependency: ACPI - Define AML device for SPCR type 0x12 devices


andreiw:
Type 0x12 can describe NS16550-like UARTs that cannot be described by PNP0501 - the e.g. register stride, clock frequency. And there are other "tunables" which may be interesting to the OS driver. This should have it's own ACPIxxxx _CID.

Can model this on https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/serial/8250.yaml.

These seem like a good set. Should these use DSD DT?

- clock-frequency
- reg-offset
- reg-shift
- reg-io-width
- fifo-size

adurbin-rivos:
'DSD DT' - what do you mean by this? device specific data and have that under a specific ACPIxxxx device name?

paul-walmsley-sifive:
> 'DSD DT' - what do you mean by this? device specific data and have that under a specific ACPIxxxx device name?

He's probably referring to the practice of embedding DT properties into the _DSD entries, i.e., https://docs.kernel.org/6.2/firmware-guide/acpi/dsd/graph.html

adurbin-rivos:
Yes, that was my question confirming if DSD DT meant that. I assumed so, but I wanted to be sure.

andreiw:
Yes, Paul is correct.

andreiw:
This would be nice to have but ideally is code-first and fueled by someone who needs it

andreiw:
Also need to add the rscv- Known Device Property Prefixex

PR #150, opened 2024-04-15
Define AML device ID/props for SPCR type 0x12 devices

andreiw:
Define AML device ID/props for SPCR type 0x12 devices

https://github.com/riscv-non-isa/riscv-brs/issues/24

### DIFF
--- a/acpi-prop.adoc
+++ b/acpi-prop.adoc
@@ -1,0 +1,25 @@
+[[acpi-props]]
+=== RVI-specific ACPI Device Properties
+
+This section defines the _DSD device properties cite:[DSD] in the `rscv-` namespace.
+
+Where explicit values are provided in a property definition, only these values must be used. System behavior with any other values is undefined.
+
+Request for additional property names in the `rscv-` namespace should be made as a git pull request to this document.
+
+[[acpi-props-uart]]
+==== Properties for UART Devices
+
+[width=100%]
+[%header, cols="10,5,25"]
+|===
+|  Property (`rscv-uart-*`) ^| Type | Description
+| `clock-frequency` | Integer | Clock feeding the IP block in Hz.
+3+| _A value of zero will preclude the ability to set the baud rate, or
+to configure a disabled device._
+| `reg-shift` | Integer | Quantity to shift the register offsets by.
+| `reg-io-width` | Integer | The size (in bytes) of the register accesses that should be performed on the device.
+3+| _1, 2, 4 or 8._
+| `rx-fifo-size` | Integer | The RX FIFO size (in bytes).
+|===
+
--- a/acpi-id.adoc
+++ b/acpi-id.adoc
@@ -9,7 +9,7 @@ consists of two parts: a Vendor ID followed by a product identifier.
 Vendor IDs consist of 4 characters, each character being either an
 uppercase letter (A-Z) or a numeral (0-9). The Vendor ID SHOULD be
 unique across the Industry and registered by the UEFI forum. For RVI
-standard devices, **"RSCV"** is the Vendor ID registered. Vendor-specific
+standard devices, `RSCV` is the Vendor ID registered. Vendor-specific
 devices can use an appropriate Vendor ID registered for the manufacturer.
 
 Product Identifiers are always four-character hexadecimal numbers (0-9
@@ -25,6 +25,8 @@ ACPI ID for any new device.
 [%header, cols="5,25"]
 |===
 | ACPI ID     ^| Device
-| RSCV0001     | RISC-V Platform-Level Interrupt Controller (PLIC)
-| RSCV0002     | RISC-V Advanced Platform-Level Interrupt Controller (APLIC)
+| `RSCV0001`     | RISC-V Platform-Level Interrupt Controller (PLIC)
+| `RSCV0002`     | RISC-V Advanced Platform-Level Interrupt Controller (APLIC)
+| `RSCV0003`     | NS16550-compatible UART compatible with a Type 0x12 SPCR definition
+2+| _Also see <<acpi-props-uart>>._
 |===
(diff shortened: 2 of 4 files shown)


leiflindholm:
I have probably missed some conversation, but how do these devices differ from a fully-compliant 16550, and how do they differ from similar components used by other architectures?

vlsunil:
Hi Andrei, I have the same question as Leif.  We use PNP0501 for standard 16550 compatible UART in ACPI (Yeah, it is giving me lot of trouble, but that's a different issue :-) ). I am not sure why we need new ACPI ID which would warrant driver changes in every OS.

andreiw:
Because PNP0501 assumes a certain reg stride (how far apart the 1 byte UART regs are... on Arm systems these were common to be on 4-byte boundaries) and I/O access capabilities (again, some Arm systems in my past only supported 4 byte accesses). Then you can have a varying RX FIFO (not detectable) and of course the bane of any UART integration - arbitrary clocking (esp for UARTs that are meant to drive high-speed blocks like BT HC)

You could say - just define the properties you want, but that will _still_ break every PNP0501 driver with such devices, because these new properties are not part of the PNP0501 contract. It's kinda like why 0x12 was added to SPCR - you can't just use 0x0000 and 0x0001, as these assume a port I/O UART _and_ the PC compatible crystal...

What happened on Arm is _everyone_ defined their own HIDs and there was no common CID. So every new platform got a new driver change with it. So this ID should make things simpler. We may add new properties in the future, but that will be okay. The contract is that anyone implementing RSCV0003 will have to keep up.

acpi-prop.adoc
```diff
@@ -0,0 +1,25 @@
...
+[width=100%]
+[%header, cols="10,5,25"]
+|===
+|  Property (`rscv-uart-*`) ^| Type | Description
+| `clock-frequency` | Integer | Clock feeding the IP block in Hz.
+3+| _A value of zero will preclude the ability to set the baud rate, or
+to configure a disabled device._
+| `reg-shift` | Integer | Quantity to shift the register offsets by.
+| `reg-io-width` | Integer | The size (in bytes) of the register accesses that should be performed on the device.
+3+| _1, 2, 4 or 8._
```
adurbin-rivos, line 22: Is the assumption that a write of a given length still only provides 1 byte of data?

andreiw: An io-width != 1 only makes sense if the reg-shift is not 1. Consider a UART with a reg-shift of 2, where it's single byte registers are aligned on 32-bit boundaries. There's been silicon that would only work (as in, the writes would only "write" something) if the accesses were also 32-bit. Such an access to a 32-bit aligned 8-bit registers would still provide only 1 byte of data, yes.

vlsunil:
> Because PNP0501 assumes a certain reg stride (how far apart the 1 byte UART regs are... on Arm systems these were common to be on 4-byte boundaries) and I/O access capabilities (again, some Arm systems in my past only supported 4 byte accesses). Then you can have a varying RX FIFO (not detectable) and of course the bane of any UART integration - arbitrary clocking (esp for UARTs that are meant to drive high-speed blocks like BT HC)
> 
> You could say - just define the properties you want, but that will _still_ break every PNP0501 driver with such devices, because these new properties are not part of the PNP0501 contract. It's kinda like why 0x12 was added to SPCR - you can't just use 0x0000 and 0x0001, as these assume a port I/O UART _and_ the PC compatible crystal...
> 
As we speak, I can see Linux PNP UART driver is getting updated to handle this case. https://www.spinics.net/lists/kernel/msg5179311.html. Atleast, linux appears to handle this with this patch series which is already ACKd. This will require ACPI to use _DSD to communicate this though. 

> What happened on Arm is _everyone_ defined their own HIDs and there was no common CID. So every new platform got a new driver change with it. So this ID should make things simpler. We may add new properties in the future, but that will be okay. The contract is that anyone implementing RSCV0003 will have to keep up.
>
I would love to have a different ACPI ID defined for standard 16550. PNP0501 is enumerated via PNP bus which has weird enumeration logic in linux causing lot of issues when we need to order its probe (i.e after interrupt controller). So, if we have ACPI ID which can be enumerated via platform bus, it will help RISC-V not to change any thing except write new platform driver (provided maintainers agree). What if same IP is used in other architectures? Can a generic CID be defined in ACPI spec itself later so that same driver written for RSCV0003 can be used?

acpi-prop.adoc
andreiw: Yep that's how NS16550 UARTs work.

andreiw:
> > Because PNP0501 assumes a certain reg stride (how far apart the 1 byte UART regs are... on Arm systems these were common to be on 4-byte boundaries) and I/O access capabilities (again, some Arm systems in my past only supported 4 byte accesses). Then you can have a varying RX FIFO (not detectable) and of course the bane of any UART integration - arbitrary clocking (esp for UARTs that are meant to drive high-speed blocks like BT HC)
> > You could say - just define the properties you want, but that will _still_ break every PNP0501 driver with such devices, because these new properties are not part of the PNP0501 contract. It's kinda like why 0x12 was added to SPCR - you can't just use 0x0000 and 0x0001, as these assume a port I/O UART _and_ the PC compatible crystal...
> 
> As we speak, I can see Linux PNP UART driver is getting updated to handle this case. https://www.spinics.net/lists/kernel/msg5179311.html. Atleast, linux appears to handle this with this patch series which is already ACKd. This will require ACPI to use _DSD to communicate this though.

RSCV0003 also relies on _DSD for properties.

> 
> > What happened on Arm is _everyone_ defined their own HIDs and there was no common CID. So every new platform got a new driver change with it. So this ID should make things simpler. We may add new properties in the future, but that will be okay. The contract is that anyone implementing RSCV0003 will have to keep up.
> 
> I would love to have a different ACPI ID defined for standard 16550. PNP0501 is enumerated via PNP bus which has weird enumeration logic in linux causing lot of issues when we need to order its probe (i.e after interrupt controller). So, if we have ACPI ID which can be enumerated via platform bus, it will help RISC-V not to change any thing except write new platform driver (provided maintainers agree). What if same IP is used in other architectures? Can a generic CID be defined in ACPI spec itself later so that same driver written for RSCV0003 can be used?

Yes, we are aligned.

PR #167, opened 2024-05-21
Remove _DEP requirement

vlsunil:
The OSPM can add IRQ dependencies by looking at GSI map. So, the requirement of adding _DEP under each device which needs GSIs (wired interrupts) is no longer required. Remove the requirement.


andreiw:
Hooray :-), I'm glad  common sense prevailed...

PR #178, opened 2024-07-22
Add initial test specification

xiaobo55x:
Add the test specification framework for BRS Spec.


andreiw:
Looks like a great start.

xiaobo55x:
> Looks like a great start.

Thanks! I will try to add more contents to it.

PR #188, opened 2024-07-30
acpi-prop: Make UART device properties to use global namespace

vlsunil:
Several Operating Systems are already using device property names for 16550-compatible UARTs in global namespace. Adding riscv- prefix will make multiple software components to change. DSD guide also allows to continue to use the same property names which are already in use. So, remove riscv-uart prefix requirement for UART devices.

Also, add reg-offset property which is also currently used by OS drivers.


vlsunil:
@andreiw @adurbin-rivos : Please review.

adurbin-rivos (approved):

PR #207, opened 2024-11-06
UIO_010 clarification attempt for https://github.com/riscv-non-isa/ri…

andreiw:
…scv-brs/issues/200

### DIFF
--- a/non-normative/uefi.adoc
+++ b/non-normative/uefi.adoc
@@ -34,6 +34,15 @@ behind a root bridge, `EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL_CONFIGURATION`
 reports valid resources assigned (e.g. for hot plug), or reports no
 resources assigned.
 
+Firmware MUST always initialize PCIe root complexes, even if booting from non-PCIe devices,
+and should not assume the OS knows how to configure root complex hardware (including,
+for example, inbound and outbound address translation windows). In fact, ECAM-compatible
+PCIe segments are assumed by operating systems to just work as per their hardware
+descriptions in ACPI and DT. Furthermore, firmware MUST perform BAR resource assignment,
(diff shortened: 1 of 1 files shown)


non-normative/uefi.adoc
```diff
@@ -34,6 +34,15 @@ behind a root bridge, `EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL_CONFIGURATION`
...
 resources assigned.
 
+Firmware MUST always initialize PCIe root complexes, even if booting from non-PCIe devices,
+and should not assume the OS knows how to configure root complex hardware (including,
+for example, inbound and outbound address translation windows). In fact, ECAM-compatible
+PCIe segments are assumed by operating systems to just work as per their hardware
+descriptions in ACPI and DT. Furthermore, firmware MUST perform BAR resource assignment,
+bridge bus number and window assignments and other reasonable device setting configuration
+(e.g. Max Payload Size) and not assume operating systems to be capable of full PCIe
+resource configuration, or to expect full reconfiguration to be necessary.
```
vlsunil, line 44: LGTM. However, when FW allocates the resources it needs to communicate to the OS that they should be preserved (via _DSM in ACPI) . Should we add that requirement as well? REF: PCI FW SPEC 4.6.5.

andreiw: Don't have a strong opinion on this one. I'm more concerned about requiring OS behavior. I think FW requesting resources to be preserved is more of a platform choice (tbh, I can only think of scenarios where the OS can't correctly reconfigure due to PCIe support being not compliant).

andreiw: Let me factor this one out into a separate issue and we can discuss that there.

vlsunil: Yeah, agreed. Thanks!

Issue #201, opened 2024-10-07
Give reasons for URT_040 and URT_050


jhauser-us:
The Architecture Review Committee would like the BRS document to include some non-normative justifications for URT_040 and URT_050, which concern the setting and persistence of non-volatile UEFI variables.

vlsunil:
@jhauser-us : Need some help understanding the feedback better.
URT_040 says - "Non-volatile UEFI variables MUST persist across calls to the `ResetSystem()` runtime service call."

It looks obvious to me that it merely states that non-volatile variables must be persistent across reboots. What type of clarification you want to see for this? 

@andreiw : Was original intent of above requirement to rule out emulated variables on RAM? If so, can we just state it explicitly that only non-volatile variables are supported? 

For URT_050 which says:
>> UEFI runtime services MUST be able to update the UEFI variables directly without the aid of an OS.

Would below additional note be sufficient to clarify?
**_UEFI variables are normally saved in a dedicated storage which is not directly accessible by the operating system._**

jhauser-us:
> It looks obvious to me that it merely states that non-volatile variables must be persistent across reboots. What type of clarification you want to see for this?

I'm not actually the source of the feedback reported in this GitHub issue; I'm just the ARC's messenger.  However, my understanding is that URT_040 is demanding that the non-volatile UEFI variables be non-volatile.  Isn't that redundant?  So why is this rule in our RISC-V BRS standard?  That sounds to me like a fair question for a non-normative comment to answer.

vlsunil:
> > It looks obvious to me that it merely states that non-volatile variables must be persistent across reboots. What type of clarification you want to see for this?
> 
> I'm not actually the source of the feedback reported in this GitHub issue; I'm just the ARC's messenger. However, my understanding is that URT_040 is demanding that the non-volatile UEFI variables be non-volatile. Isn't that redundant? So why is this rule in our RISC-V BRS standard? That sounds to me like a fair question for a non-normative comment to answer.

Yeah. it looks redundant to say so. But I think this requirement just exists to prohibit volatile UEFI variables emulated in RAM which then requires an additional sentence to say so actually. I hope @andreiw or @adurbin-rivos can clarify.

adurbin-rivos:
I can't speak to the requirement on the variables. I agree that it seems redundant. Likewise, I think the note "UEFI variables are normally saved in a dedicated storage which is not directly accessible by the operating system." seems fine as well.

[...]

Issue #198, opened 2024-09-28
Despite claim, no properties are defined in the rscv- namespace


jhauser-us:
Section 6.3, "RVI-specific ACPI Device Properties", says, "This section defines the `_DSD` device properties in the `rscv-` namespace."  However, the only properties actually mentioned in this section are for UART devices, and if I'm not mistaken, these are *not* defined in the `rscv-` namespace.  (Right?)

In short, I believe this section says it defines properties in the `rscv-` namespace but then doesn't do so.

vlsunil:
You are right. I think this is a place holder for any property to be defined under rscv- namespace. The DSD guide (https://github.com/UEFI/DSD-Guide/blob/main/src/dsd-guide.adoc#appendix-c-known-device-property-prefixes) allows to use vendor ACPI ID as the prefix for the property but vendor needs to manage those properties. For RISC-V, ACPI ID registered is RSCV. Hence, any one in future wants to define and use a new property under rscv- namespace, should register here first. I think that is the idea to have this section.

andreiw:
We had some properties here initially, then - we didn't. Nevertheless, it makes sense to stake a claim to the rscv- namespace now.

andreiw:
Marking this as won't fix given last two comments. If you're okay with that, please close the ticket @jhauser-us.

jhauser-us:
The right response is not to leave the document making an obviously false statement.  If you want to stake the claim to the `rscv-` namespace, then change the text to say that.

jhauser-us:
It is the consensus of the Architecture Review Committee that Section 6.3, "RVI-specific ACPI Device Properties", needs a non-normative note clarifying that the table of `rscv-` names is currently empty.  And this paragraph:
> Request for new property names in the `rscv-` namespace should be made as a git pull request to below table in this document.

should be made non-normative, preferably moved into the same non-normative note.

In my opinion, it makes more sense for this non-normative note to *follow* the table, which would require "below table" be changed appropriately.  (Note, by the way, that the words *above* and *below* are not adjectives, so it's not gramatically correct to say "above table" or "below table".)

vlsunil:
Thanks @jhauser-us . Let me update as per your suggestion.

Issue #182, opened 2024-07-30
satp-specified address translation cannot be disabled


jhauser-us:
For UEFI_030, what does "Enable address translation" mean?

Technically, when S-mode is implemented, the RISC-V Privileged Architecture does not provide any way to disable address translation for privilege modes less-privileged than M-mode.  Is this requirement trying to say that the address translation mode configured in register `satp` cannot be Bare?  If so, why is Bare excluded?

And how is UEFI_030 expected to square with the EBBR requirement (relevant for BRS-B) that, for UEFI Boot Services, "All RAM defined by the UEFI memory map must be identity-mapped, which means that virtual addresses must equal physical addresses"?  Is UEFI_030's "Enable address translation" (whatever it means) only intended to apply after the call to `ExitBootServices`?

andreiw:
This means that UEFI itself is making use of MMU facilities. It still must meet the requirement that all RAM is 1:1 mapped, of course - virtual addresses equaling physical addresses. But bare is not allowed.

Using the MMU is important for:
a) no-execute protecting data regions, stacks, etc
b) no-execute protecting non-native code regions when running x86 OpRoms (e.g. to detect transitions into non-native code)
c) correct memory attributes for device DMA when Svpbmt is available, where necessary

andreiw:
Please close if no further action is desired.

jhauser-us:
> This means that UEFI itself is making use of MMU facilities.  [...]

Okay, then there are some things about this rule that I really don't understand.  After booting, how does UEFI modify the page tables that are exclusively under the control of the OS?  Or, if the UEFI doesn't modify the page tables itself, then how does it enforce the page-based protections that you say it requires/expects?  If the OS is responsible for providing the protections that UEFI needs/expects, then "Enable address translation" doesn't cover it.

Is "Enable address translation" only meant to apply before `EFI_EVENT_GROUP_READY_TO_BOOT`?  That's not what it says.  And even so, it still seems to me that "Enable address translation" by itself doesn't guarantee the protections you listed.  The page tables must be properly configured for that outcome, but there's no mention of that part.  Are we assuming readers will understand that's what this requirement is about?  I'm not confident of that.  Just saying "address translation" doesn't, in my opinion, get the point across, especially as there's actually an EBBR requirement (maybe expected for RISC-V UEFI generally?) that addresses *not* be interestingly translated (identity translation only).

What else am I missing?

[...]

PR #208, opened 2024-11-08
Updates for ARC feedback #182, #198 and #201

vlsunil:
My attempt to resolve ARC feedback regarding UEFI variable persistence, SATP and ACPI device property section.  Please review and suggest better wordings or if I am missing any thing.

### DIFF
--- a/uefi.adoc
+++ b/uefi.adoc
@@ -19,8 +19,9 @@ IMPORTANT: All content in this section is optional and recommended for BRS-B.
 | `UEFI_030` a| MUST meet the following memory map rules:
 
               * The default memory space attribute is `EFI_MEMORY_WB`.
-              * Address translation is enabled.
+              * Paged virtual-memory scheme MUST be configured by the firmware with identity mapping and MUST support `EFI_MEMORY_ATTRIBUTE_PROTOCOL` protocol.
               * Only use `EfiRuntimeServicesData` memory type for describing any SMBIOS data structures.
+2+| _Enabling paged virtual memory scheme means, `satp.MODE` must not be set to `Bare` and proper page table with identity mapping is setup by the firmware. This is required for platform protection use cases before handing off to the OS._
 | `UEFI_040` | An implementation MAY comply with the _UEFI Platform Initialization Specification_ cite:[UEFI-PI].
 | `UEFI_050` | All hart manipulation internal to a firmware implementation SHOULD be done before completion of the `EFI_EVENT_GROUP_READY_TO_BOOT` event. Firmware MUST place all secondary harts in an offline state before completion of the EFI_EVENT_GROUP_READY_TO_BOOT event.
 2+| _This ensures an OS loader is entered with an OS-compatible state for all harts.The OS loader and/or the OS may resume the secondary harts, if required, as part of their boot and join sequence._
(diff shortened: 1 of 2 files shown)


vlsunil:
@andreiw @jhauser-us @avpatel @adurbin-rivos - please review.

andreiw (commented):
The intention of URT_040 was to avoid RPi-like UEFI implementations, where NV variables are in practice volatile after ExitBootServices. This leads to bad experiences with OS installers and probably other things.

Pi 4 had no non-volatile store beyond the SD card. Variables were persisted in RPI_EFI.FD during Boot Services, but he SD card could not be accessed after OS use, so - no non-volatile vars in the OS.

URT_040 refers to a behavior wrt ResetSystem, and that was intended - again, deficient designs like RPi4 could, during their ResetSystem path, flush out NV variables to persistent store (now that the OS is out of the way)

andreiw (commented):
I hate to be pendantic, but the UEFI spec *already* talks about identity mapping for RAM regions as a requirement.

Probably want an additional requirement that EFI_MEMORY_ATTRIBUTE_PROTOCOL and EFI_CPU_ARCH_PROTOCOL.SetMemoryAttributes() are implemented and non-dummy (really affects page attributes).

adurbin-rivos:
It seems that people feel that system designs should follow the already written requirements in the specs vs trying to re-describe or over-describe things to keep people from shooting themselves in the foot. I'm fine w/ these changes as I'm more on the side of "people should design things correctly" vs trying to prevent larger mistakes.

adurbin-rivos (approved):

ved-rivos:
If its a common mistake then its not a bad idea to call it out. The rule book is intended to be helpful.

uefi.adoc
```diff
@@ -19,8 +19,9 @@ IMPORTANT: All content in this section is optional and recommended for BRS-B.
 | `UEFI_030` a| MUST meet the following memory map rules:
 
               * The default memory space attribute is `EFI_MEMORY_WB`.
-              * Address translation is enabled.
+              * Paged virtual-memory scheme MUST be configured with identity mapping to support memory protection.
               * Only use `EfiRuntimeServicesData` memory type for describing any SMBIOS data structures.
+2+| _Enabling paged virtual memory scheme means satp.MODE must not be set to `Bare` and proper page table for identity mapping is setup by the firmware. This is required for platform protection use cases before handing off to the OS._
```
avpatel, line 24: s/proper page table for identity mapping/proper page table with identity mapping/

I think the page table is setup by OS and this page table identity mapping is required when OS makes EFI call at boot-time or run-time.

vlsunil:
> The intention of URT_040 was to avoid RPi-like UEFI implementations, where NV variables are in practice volatile after ExitBootServices. This leads to bad experiences with OS installers and probably other things.
> 
> Pi 4 had no non-volatile store beyond the SD card. Variables were persisted in RPI_EFI.FD during Boot Services, but he SD card could not be accessed after OS use, so - no non-volatile vars in the OS.
> 
> URT_040 refers to a behavior wrt ResetSystem, and that was intended - again, deficient designs like RPi4 could, during their ResetSystem path, flush out NV variables to persistent store (now that the OS is out of the way)

Ahh OK. Thanks!. I thought that was the intention. In that case, let's rephrase the requirement that "The UEFI variables MUST be persistent across ResetSystem() call".  I hope that should be clear enough requirement.

vlsunil:
> I hate to be pendantic, but the UEFI spec _already_ talks about identity mapping for RAM regions as a requirement.
>
While I agree with that, I think it is better to mention again here since we are talking about MMU configuration. This should make it clear that we are still requiring identity map while configuring non-BARE mode.

> Probably want an additional requirement that EFI_MEMORY_ATTRIBUTE_PROTOCOL and EFI_CPU_ARCH_PROTOCOL.SetMemoryAttributes() are implemented and non-dummy (really affects page attributes).

Good point. However, it is better to avoid requiring EFI_CPU_ARCH_PROTOCOL which is part of PI spec. Better to leave to the FW implementation to provide EFI_MEMORY_ATTRIBUTE_PROTOCOL even when they are not PI complaint. WDYT?

uefi.adoc
vlsunil: > s/proper page table for identity mapping/proper page table with identity mapping/
> 
Thanks!. Let me update.

xypron:
> In that case, let's rephrase the requirement that "The UEFI variables MUST be persistent across ResetSystem() call". I hope that should be clear enough requirement.

This get's it wrong. Volatile variables should not be persisted. Only variables with attribute EFI_VARIABLE_NON_VOLATILE must be persisted.

PR #211, opened 2024-12-07
Minor updates

vlsunil:
1) One more place where change to "rule" from "requirement" was pending pointed by John.
2) A stale note about PMU extension being ratified is removed since it is ratified now.


adurbin-rivos (approved):

avpatel:
Looks good to me as well.

PR #225, opened 2025-02-06
Minor updates before freeze

vlsunil:
1) Reserve ACPI IDs for RPMI/MPXY based system MSI controller and mailbox. Not functional / requirement changed.
2) Move ACPI_080 under AML requirements. No change in requirements, just reorg.
3) Move URT_060 under security requirements.
4) Update the version as per new RVI versioning scheme.


vlsunil:
@avpatel - please review.

PR #231, opened 2025-03-18
Update BRS / TS revision and date for freeze.

jamieiles:
Prepare for public review as per https://lf-riscv.atlassian.net/browse/RVS-2155

### DIFF
--- a/header.adoc
+++ b/header.adoc
@@ -2,9 +2,9 @@
 :description: RISC-V Boot and Runtime Services Specification Document (BRS)
 :company: RISC-V.org
 :url-riscv: http://riscv.org
-:revdate: 7th February 2025
-:revnumber: 0.8
-:revremark: This document is Stable.
+:revdate: 18th March 2025
+:revnumber: 0.9
+:revremark: This document is Frozen.
 :doctype: book
 :preface-title: Preamble
 :colophon:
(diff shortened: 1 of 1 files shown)


brs_ts_header.adoc
```diff
@@ -2,9 +2,9 @@
 :description: RISC-V Boot and Runtime Services Test Specification
 :company: RISC-V.org
 :url-riscv: http://riscv.org
-:revdate: 7th February 2024
-:revnumber: 0.1
-:revremark: This document is in Development state.  Change should be expected.
+:revdate: 18th March 2025
+:revnumber: 0.9
+:revremark: This document is in Frozen state.  Change is extremely unlikely.
```
vlsunil, line 7: The test spec is incomplete and it is not part of the ratification process. So, it is intentionally kept as development version.

header.adoc
```diff
@@ -39,9 +39,9 @@ BRS Task Group
 
 // Preamble
 [WARNING]
-.This document is in the link:https://lf-riscv.atlassian.net/wiki/display/HOME/Specification+States[Stable state]
+.This document is in the link:https://lf-riscv.atlassian.net/wiki/display/HOME/Specification+States[Frozen state]
 ====
-Assume anything could still change, but limited change should be expected.
+Change is extremely unlikely.
```
vlsunil, line 44: Thanks!. But as per official documentation template (https://github.com/riscv/docs-resources?tab=readme-ov-file#document-state-markings), it should be like below.

[WARNING]
.This document is in the link:https://lf-riscv.atlassian.net/wiki/display/HOME/Specification+States[Frozen state]
====
Change is extremely unlikely. A high threshold will be used, and a change
will only occur because of some truly critical issue being identified during
the public review cycle. Any other desired or needed changes can be the subject
of a follow-on new extension.
====

brs_ts_header.adoc
jamieiles: Thanks, I've removed this

header.adoc
jamieiles: Updated from the template, thanks

vlsunil (approved):
I will update the revdate separately to reflect latest.

PR #232, opened 2025-03-19
uefi: add URT_060 to support Landing Pads in UEFI Runtime Services

radimkrcmar:
The caller of UEFI runtime services might be running with enabled Landing Pads.  Require support for Landing pads in UEFI Runtime Services to avoid disabling the protection before every call.

I did not mention Landing pads in the spec, because there might be future extensions that do something similar, and we should support them all.

I would appreciate suggestions that improve the wording, because the required bit is actually in the EFI_MEMORY_ATTRIBUTES_TABLE, thanks.

### DIFF
--- a/uefi.adoc
+++ b/uefi.adoc
@@ -70,6 +70,8 @@ hand-off info to an OS, for example, to provide RAM disk information
 2+| _This rule is included in this specification to address a common mistake in implementing the UEFI requirements for non-volatile variables, even though it may appear redundant with the existing UEFI specification._
 | `URT_050` | UEFI runtime services MUST be able to update the UEFI variables directly without the aid of an OS.
 2+| _UEFI variables are normally saved in a dedicated storage which is not directly accessible by the operating system._
+| `URT_060` | If UEFI runtime services can be called with enabled forward control flow guard, then UEFI runtime services MUST have the `EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD` bit set to 1.
+2+| _This rule allows calls to UEFI runtime services without disabling the forward control flow guard on platforms that allow the OS to enable it._
 |===
 
 === BRS-I Security Requirements


uefi.adoc
```diff
@@ -70,6 +70,8 @@ hand-off info to an OS, for example, to provide RAM disk information
 2+| _This rule is included in this specification to address a common mistake in implementing the UEFI requirements for non-volatile variables, even though it may appear redundant with the existing UEFI specification._
 | `URT_050` | UEFI runtime services MUST be able to update the UEFI variables directly without the aid of an OS.
 2+| _UEFI variables are normally saved in a dedicated storage which is not directly accessible by the operating system._
+| `URT_060` | If UEFI runtime services can be called with enabled forward control flow guard, then UEFI runtime services MUST have the `EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD` bit set to 1.
```
adurbin-rivos, line 73: I have a few questions on the intent of using the flag in general:

- How would UEFI know to set this flag? 
- And for what memory regions?

adurbin-rivos: I guess that flag field applies to all EFI_MEMORY_DESCRIPTOR in the EFI_MEMORY_DESCRIPTOR. And what it signals is that the UEFI was built with the landing pads where the OS could take advantage?

radimkrcmar: Exactly.  I haven't spent much time in the UEFI space, so the wording could probably be clearer.

ved-rivos:
In general, BRS should not have to repeat what UEFI says . Is there some way BRS needs to modulate or override the UEFI rules that make this addition necessary

uefi.adoc
adurbin-rivos: My critique to the wording would be the following: "then UEFI runtime services MUST"  It's not a property of the runtime services being marked with that flag. It's the EFI_MEMORY_ATTRIBUTES_TABLE. While I don't think that makes logical sense, that's how the spec was written.

adurbin-rivos:
> In general, BRS should not have to repeat what UEFI says . Is there some way BRS needs to modulate or override the UEFI rules that make this addition necessary

FWIW: I see this definition in the spec:

`#define EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD  0x1
// BIT0 implies that Runtime code includes the forward control flow guard
//      instruction, such as X86 CET-IBT or ARM BTI.`

I don't think it says what to do in the spec aside from explaining what the flag means. I didn't search profusely, though. I was relying on the search engine to pull that flag up.

radimkrcmar:
> I don't think it says what to do in the spec aside from explaining what the flag means. I didn't search profusely, though. I was relying on the search engine to pull that flag up.

Yes, as I understand it, the edk2 implementation doesn't set the bit when it loads runtime services that aren't compiled with forward control flow:
```
  //
  // Check whether we are loading a runtime image that lacks support for
  // IBT/BTI landing pads.
  //
  if ((Image->ImageContext.ImageCodeMemoryType == EfiRuntimeServicesCode) &&
      ((Image->ImageContext.DllCharacteristicsEx & EFI_IMAGE_DLLCHARACTERISTICS_EX_FORWARD_CFI_COMPAT) == 0))
  {
    gMemoryAttributesTableForwardCfi = FALSE;
  }
```
```
  if (gMemoryAttributesTableForwardCfi) {
    MemoryAttributesTable->Flags = EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD;
  } else {
    MemoryAttributesTable->Flags = 0;
  }
```

uefi.adoc
ved-rivos: UEFI cannot predict whether "can be called with enabled forward control flow ". It can only declare what it has - it either sets the flag or does not based on how its compiled. Its for the caller to use this attribute to determine how to make the call i.e. make the call with forward control flow integrity enabled or to make the call with forward control flow integrity disabled.

radimkrcmar: Yes, if UEFI doesn't know what platforms it will execute on, then this rule would force it to be compiled with FCFI, because some caller might do that.

Currently, landing pads are an optional extension and can only be enabled by higher-privileged mode.
I was trying to give some freedom to UEFI that is tightly coupled with the platform -- it can know whether the platform allows S-mode to execute with landing pads.  If the platform doesn't allow that, then there is no point to force UEFI runtime services to be compiled with landing pads.

radimkrcmar: The general goal is to simplify the call, because it's much easier to have landing pads in UEFI runtime services than to toggle landing pads.
Is this too constraining for BRS?   I'm thinking we could do that in the Server Platform as well.

ved-rivos: While landing pads are optional, compiling with landing pads is always benign since they become NOP when the implementation does not support landing pads. So this rule could be stated one of two ways:

"UEFI firmware MUST be compiled with Zicfilp and MUST set EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD"

or

"If UEFI firmware is compiled with Zicfilp then it MUST set EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD"

The second version is redundant since it is stating what UEFI already says - so not sure it needs to be in BRS.

The first is a mandate which would be unique and goes over and above what UEFI says and would be appropriate an appropriate rule. However, then it may start contradicting with BRS 1.0 which claims that it only needs to be compatible with RVA20 or higher. This rule would then want to be in the server platform specification. 

The platform does not allow or disallow use of landing pads by S-mode. If the OS has been compiled with landing pads and the policy set by the OS administrator requires it to be enabled and Zicfilp is supported then the OS will make SBI call to enable it. If the OS has enabled landing pads for itself it can further discover whether the UEFI code has landing pads and then decide how to call the UEFI code.

radimkrcmar: > While landing pads are optional, compiling with landing pads is always benign since they become NOP when the implementation does not support landing pads. So this rule could be stated one of two ways:
>
> "UEFI firmware MUST be compiled with Zicfilp and MUST set EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD"
>
> "If UEFI firmware is compiled with Zicfilp then it MUST set EFI_MEMORY_ATTRIBUTES_FLAGS_RT_FORWARD_CONTROL_FLOW_GUARD"
> 
> The second version is redundant since it is stating what UEFI already says - so not sure it needs to be in BRS.

I agree, the second version does nothing and we shouldn't have it.

> The first is a mandate which would be unique and goes over and above what UEFI says and would be appropriate an appropriate rule. However, then it may start contradicting with BRS 1.0 which claims that it only needs to be compatible with RVA20 or higher. This rule would then want to be in the server platform specification.

The first version excludes desirable UEFI implementations to make the rule simpler, but it still doesn't break anything on RVA20, so maybe we're fine here.

> The platform does not allow or disallow use of landing pads by S-mode. If the OS has been compiled with landing pads and the policy set by the OS administrator requires it to be enabled and Zicfilp is supported then the OS will make SBI call to enable it. If the OS has enabled landing pads for itself it can further discover whether the UEFI code has landing pads and then decide how to call the UEFI code.

I wanted to convey two ideas when saying that the platform might not allow lading pads:
* Zicfilp might not be supported.
* The SBI call might not succeed even if Zicfilp is supported.  (afaik, it's not mandated)

To expand on this, I have several design goals for the rule:

1. Do not force the OS to have code that disables landing pads when calling UEFI runtime services (it could be leveraged as a gadget in an attack).
2. Allow the OS to use the SBI to lock enabled landing pads and still use UEFI runtime services.
3. Do not force UEFI to compile all code with landings pads.
4. Do not force UEFI to compile runtime services with landing pads on machines that don't have Zicfilp.
5. Do not force UEFI to compile runtime services with landing pads when SBI doesn't allow landing pads for the OS.
6. Do not force UEFI to add the flag to unrelated memory descriptors.
7. Make the rule compatible with all possible future extensions that fall under the UEFI flag.

The first version doesn't achieve (3), (4), (5), (6), and (7).  The original wording is more complex, but for a reason.

vlsunil: I would vote for having this requirement in server platform spec first and then bring it to BRS 2.0 when its minimum support would be (?) RVA23 profile. I like Ved's first version to add in server platform spec. I don't think it is a good idea to make a condition on firmware build that the OS is capable and enabled a feature. So, keeping the requirement in platform specs ensures hardware is capable which should be sufficient for FW to be compiled with Zicfilp. This should solve (4).

MAT is only for RuntimeSevices(Code and Data). So, I think that should resolve (3).  I am not sure what does (6) mean given MAT is for Runtime descriptors only. I also can not tell about (7) since we can't predict the future. We can update UEFI spec itself if required in future if the current flag is not sufficient.

radimkrcmar: > I would vote for having this requirement in server platform spec first and then bring it to BRS 2.0 when its minimum support would be (?) RVA23 profile. I like Ved's first version to add in server platform spec.

Ok, I'll prepare a PR against the server platform.

> I don't think it is a good idea to make a condition on firmware build that the OS is capable and enabled a feature. So, keeping the requirement in platform specs ensures hardware is capable which should be sufficient for FW to be compiled with Zicfilp. This should solve (4).

Sounds good.  The condition also covers (5), and to remove the condition here, we can add another platform requirement: SBI must allow the call to enable landing pads.

> MAT is only for RuntimeSevices(Code and Data). So, I think that should resolve (3).

We still want to say that only Runtime services must be compiled with Zicfilp.
Other UEFI code can do what it wants.

> I am not sure what does (6) mean given MAT is for Runtime descriptors only.

I wasn't sure of that, thanks.  I won't add a specification of which MAT it is, then.

> I also can not tell about (7) since we can't predict the future. We can update UEFI spec itself if required in future if the current flag is not sufficient.

No, we can only prepare for all possibilities.  I meant to address other way around, when RISC-V has another extension that falls under the UEFI flag.  i.e. say FCFI instead of Zicfilp.
The generalization might be unnecessary, but I think that it doesn't hurt us now and could prevent some pain in the future.

vlsunil:
This will be added in server platform spec. We can revisit whether it makes sense to bring it in BRS 2.0. Thanks!

PR #247, opened 2025-08-29
Update version and state to reflect ratified.

vlsunil:
BoD has ratified the BRS spec. Update the version and state of the document and release 1.0.


avpatel:
Looks good to me.

PR #282, opened 2025-12-24
Fix SBI_070 rule

wxjstz:
The EDK2 includes code to detect extension support, functioning when console putchar and console getchar are supported.


vlsunil:
Legacy extensions are deprecated in favor of later extensions. Having edk2 code supporting either doesn't mean spec also  should support. DBCN is the forward looking requirement. So, I don't see any need to support legacy extension in the spec.

